PR #13, opened 2020-06-03
chore(package.json): using local docsify instead of npx

willmendesneto:
Removing npx call for docsify and using local package instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "build:esm": "babel src --out-dir build/esm --extensions \".ts,.tsx\" --source-maps inline --ignore src/__tests__",
     "build:types-cjs": "tsc --emitDeclarationOnly --project tsconfig.build.json --outDir build/cjs && cp src/index.js.flow build/cjs",
     "build:types-esm": "tsc --emitDeclarationOnly --project tsconfig.build.json --outDir build/esm && cp src/index.js.flow build/esm",
-    "docs": "npx docsify-cli serve docs",
+    "docs": "docsify serve docs",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -44,7 +44,7 @@
     "build:esm": "babel src --out-dir build/esm --extensions \".ts,.tsx\" --source-maps inline --ignore src/__tests__",
     "build:types-cjs": "tsc --emitDeclarationOnly --project tsconfig.build.json --outDir build/cjs && cp src/index.js.flow build/cjs",
     "build:types-esm": "tsc --emitDeclarationOnly --project tsconfig.build.json --outDir build/esm && cp src/index.js.flow build/esm",
-    "docs": "npx docsify-cli serve docs",
+    "docs": "docsify serve docs",
```
rohan-deshpande, line 47: Ah okay so the command should just be `docsify` and not `docsify-cli` if you've got it locally installed? I actually wanted to remove it from the dev dependencies but if this works then that's fine too.

willmendesneto: Yeah, actually it's added as a devDependency already. So I'll leave for you to decide, in that case

rohan-deshpande:
@willmendesneto so I remember why we chose to use npx now. Docsify has a security vulnerability in it that hasn't been fixed. Better to remove it from dev dependencies and leave the npx call for now.

package.json
rohan-deshpande: https://github.com/atlassian-labs/react-resource-router/pull/13#issuecomment-638653545

willmendesneto:
All good! Closing this pull request since it's not helpful. Thanks!

PR #21, opened 2020-06-11
Use type keyword for type exports

rohan-deshpande:
### Fixed 

* Type exports no longer make it into the `build/esm/index.js` file fixing a bug noticed when using the library in CRA apps

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -22,7 +22,7 @@ export {
   getRouteContext,
 } from './common/utils';
 
-export {
+export type {
   Location,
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -22,7 +22,7 @@ export {
   getRouteContext,
```
MonicaOlejniczak, line 22: Looks like you missed a case in the `MemoryRouter`since `MemoryRouterProps` is exported in the index.tsx file.

PR #24, opened 2020-06-12
fix withRouter types

billyjs:
This was previously merged into @atlaskit/router on atlassian-frontend for AKR-23 but didn't get migrated to RRR.

### DIFF
--- a/src/common/types.ts
+++ b/src/common/types.ts
@@ -110,12 +110,16 @@ export type RouterStoreContext = {
   location: Location;
 };
 
+type RouteComponentProps = Partial<RouteContext> & {
+  [k: string]: any;
+};
+
 export type Route = {
   path: string;
   exact?: boolean;
 
   /** The component to render on match, typed explicitly */
-  component: ComponentType<RouteContext>;
+  component: ComponentType<RouteComponentProps> | ComponentType<any>;
 
   /** If present, must return true to include the route. */
   enabled?: () => boolean;
--- a/src/controllers/with-router/index.tsx
+++ b/src/controllers/with-router/index.tsx
@@ -1,10 +1,14 @@
 import React, { ComponentType } from 'react';
 
 import { BrowserHistory, RouteContext } from '../../common/types';
+import { RouterActionPush, RouterActionReplace } from '../router-store/types';
 import { RouterSubscriber } from '../subscribers/route';
 
-// TODO
-type WithRouter = RouteContext & { history: BrowserHistory };
+export type WithRouterProps = RouteContext & {
+  history: BrowserHistory;
+  push: RouterActionPush;
+  replace: RouterActionReplace;
+};
 
 const getWrappedComponentDisplayName = (
   component: ComponentType<any>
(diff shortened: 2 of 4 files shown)


src/controllers/with-router/index.tsx
```diff
@@ -23,20 +27,20 @@ const getWrappedComponentDisplayName = (
...
-export const withRouter = <P extends Record<string, any>>(
+export const withRouter = <P extends Record<string, any> = {}>(
   WrappedComponent: ComponentType<P>
-) => {
+): ComponentType<Omit<P, keyof WithRouterProps>> => {
   const displayName = getWrappedComponentDisplayName(WrappedComponent);
-  const Component = WrappedComponent as ComponentType<WithRouter & P>;
-  const ComponentWithRouter = (props: P) => (
+  const Component = WrappedComponent;
+  const ComponentWithRouter = (unknownProps: any) => (
```
rohan-deshpande, line 35: is "unknown" necessary?

billyjs: no it's not

rohan-deshpande: I would just leave it as `props` if possible

src/controllers/with-router/index.tsx
```diff
@@ -23,12 +27,12 @@ const getWrappedComponentDisplayName = (
...
 
-export const withRouter = <P extends Record<string, any>>(
+export const withRouter = <P extends Record<string, any> = {}>(
   WrappedComponent: ComponentType<P>
-) => {
+): ComponentType<Omit<P, keyof WithRouterProps>> => {
   const displayName = getWrappedComponentDisplayName(WrappedComponent);
-  const Component = WrappedComponent as ComponentType<WithRouter & P>;
-  const ComponentWithRouter = (props: P) => (
+  const Component = WrappedComponent;
```
MonicaOlejniczak, line 34: /nit May as well just renamed the `WrappedComponent` argument to `Component`?

src/controllers/with-router/index.tsx
```diff
@@ -23,12 +27,12 @@ const getWrappedComponentDisplayName = (
...
+  const ComponentWithRouter = (props: any) => (
```
MonicaOlejniczak, line 35: Why can't you use `P` anymore over `any`?

src/common/types.ts
```diff
@@ -110,12 +110,16 @@ export type RouterStoreContext = {
...
+  [k: string]: any;
+};
+
 export type Route = {
   path: string;
   exact?: boolean;
 
   /** The component to render on match, typed explicitly */
-  component: ComponentType<RouteContext>;
+  component: ComponentType<RouteComponentProps> | ComponentType<any>;
```
MonicaOlejniczak, line 122: Could this just be made into an optional generic like in the `withRouter` case? This should only need to be supported for consumers that do not use the exported `RouteComponent` component, and I would avoid the use of `any` here as a result. The default prop type just just be `RouteComponentProps` for those using the route component.

PR #25, opened 2020-06-24
Added universal-router controller

rohan-deshpande:
### Added 

* `UniversalRouter` component - this component should work universally in both the browser and the server. 
* Tests for this which include all unit tests from the three routers we currently have 

### Fixed 

* The need for weird app tree composition described here https://atlassian-labs.github.io/react-resource-router/#/router/ssr?id=tree-composition

### Note 

* If we merge this and are happy with it, we can eventually delete `Router`, `StaticRouter` and `MemoryRouter` and just have this code as the only router that needs to be used.

### DIFF
--- a/src/common/utils/is-node-environment/index.ts
+++ b/src/common/utils/is-node-environment/index.ts
@@ -1,0 +1,11 @@
+export const isNodeEnvironment = () => {
+  if (typeof window === 'undefined') {
+    return true;
+  }
+
+  if (window.name === 'nodejs') {
+    return true;
+  }
+
+  return false;
+};
--- a/src/common/utils/is-node-environment/index.ts
+++ b/src/common/utils/is-node-environment/index.ts
@@ -1,9 +1,18 @@
+/**
+ * @see https://github.com/jsdom/jsdom/releases/tag/12.0.0
+ * @see https://github.com/jsdom/jsdom/issues/1537
+ */
+const isJsDomEnvironment = () =>
+  window.name === 'nodejs' ||
+  navigator.userAgent.includes('Node.js') ||
+  navigator.userAgent.includes('jsdom');
+
 export const isNodeEnvironment = () => {
   if (typeof window === 'undefined') {
     return true;
   }
 
-  if (window.name === 'nodejs') {
+  if (isJsDomEnvironment()) {
     return true;
   }
 
--- a/src/controllers/universal-router/index.tsx
+++ b/src/controllers/universal-router/index.tsx
@@ -19,7 +19,7 @@ const getInferredHistory = (history: BrowserHistory, location?: string) =>
  */
 export class UniversalRouter extends Component<UniversalRouterProps> {
   static defaultProps = {
-    isStatic: false,
+    isGlobal: true,
     history: DEFAULT_HISTORY,
   };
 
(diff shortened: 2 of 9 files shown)


src/controllers/universal-router/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+    } = this.props;
+
+    return (
+      <RouterContainer
+        routes={routes}
+        history={getInferredHistory(history, location)}
+        isStatic={getIsStaticRouter(isStatic)}
+        resourceContext={resourceContext}
+        resourceData={resourceData}
+        isGlobal
```
albertogasparin, line 96: Can we take the opportunity of making `isGlobal` "prop driven"? Especially because in tests global complicates things.
I think the default can still be `true` but we could allow it to be set to false explicitly

src/controllers/universal-router/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+const getInferredHistory = (history: BrowserHistory, location?: string) =>
+  location ? createMemoryHistory({ initialEntries: [location] }) : history;
+
+/**
+ * Default prop provider for the RouterContainer.
+ *
+ */
+export class UniversalRouter extends Component<UniversalRouterProps> {
+  static defaultProps = {
+    isStatic: false,
```
albertogasparin, line 25: Can you explain again what is `isStatic` for?

rohan-deshpande: The prop prevents the router from listening to history. We don’t want or need the router to listen to history in a server context. Perhaps we can fully replace it with the `isNodeEnvironment()` util but let me look into it more and see if there’s any other reason we’d want an explicit prop.

src/common/utils/is-node-environment/index.ts
```diff
@@ -0,0 +1,11 @@
+export const isNodeEnvironment = () => {
+  if (typeof window === 'undefined') {
+    return true;
+  }
+
+  if (window.name === 'nodejs') {
```
jackbrown, line 6: JSDOM doesn't set `window.name` to `nodejs` as of version 12 😕

rohan-deshpande: @albertogasparin this will be a problem in RLL as well then, so we need a fix, this issue https://github.com/jsdom/jsdom/issues/1537 states that we should use 

```
navigator.userAgent.includes("Node.js") || navigator.userAgent.includes("jsdom")
```

But I think this is so jsdom specific... it depends on whether we are comfortable saying that on the server jsdom is a hard dependency.

rohan-deshpande: We could potentially use https://github.com/flexdinesh/browser-or-node

src/common/utils/is-node-environment/index.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * @see https://github.com/jsdom/jsdom/releases/tag/12.0.0
+ * @see https://github.com/jsdom/jsdom/issues/1537
+ */
+const isJsDomEnvironment = () =>
```
jackbrown, line 5: I feel it might be better to just check if we're rendering in node, rather than checking for jsdom. Some users might be rendering without jsdom which would fail this check.

rohan-deshpande: I agree it's quite jsdom centric. I think we still need the jsdom check in here, but perhaps the check for node can not rely on just checking window.

rohan-deshpande: Updated the function

PR #28, opened 2020-06-28
Bump sweet-state and fix tests

albertogasparin:


src/__tests__/unit/controllers/redirect/test.tsx
```diff
@@ -20,7 +20,7 @@ const MockHistory = {
   goForward: jest.fn(),
   block: jest.fn(),
   registerBlock: jest.fn(),
-  listen: jest.fn(),
```
albertogasparin, line 23: The listen mock was wrong as it has to return an unsubscribe function otherwise `RouterSubscriber` will call `listen()` action at every render and cause a render loop. Discovered that now as prev sweet-state version was over-optimising subscribers, memoizing even when not using selectors (but the hook version didn't) but it was only by luck.

src/__tests__/unit/controllers/router-store/test.tsx
```diff
@@ -615,9 +615,6 @@ describe('SPA Router store', () => {
       expect(componentRenderStates).toEqual([
         'loading:A',
         `data:${dataA1}`,
-        // this duplicate render state required in this test as of react-sweet-state@2.2.0
-        // @see https://github.com/atlassian/react-sweet-state/issues/81
-        `data:${dataA1}`,
```
albertogasparin, line 620: This is what we were waiting for, the actual fix for unnecessary re-renders  :D

PR #29, opened 2020-06-28
Remove default legacy history and fix history on render

albertogasparin:
- `UniversalRouter` should not consume default history anymore, as we want to deprecate such behaviour
- `UniversalRouter` was creating new history at every re-render

### DIFF
--- a/src/__tests__/unit/controllers/universal-router/test.tsx
+++ b/src/__tests__/unit/controllers/universal-router/test.tsx
@@ -126,7 +126,7 @@ describe('UniversalRouter', () => {
         .mockImplementation(() => mockHistory);
     });
 
-    it('should not respond to history changes', () => {
+    it('should listen to memory history if no history provided', () => {
       mount(
         <Router routes={[]} isGlobal={false}>
           <RouterSubscriber>
@@ -135,7 +135,7 @@ describe('UniversalRouter', () => {
         </Router>
       );
 
-      expect(mockHistory.listen).not.toHaveBeenCalled();
+      expect(mockHistory.listen).toHaveBeenCalled();
     });
 
     describe('static requestResources', () => {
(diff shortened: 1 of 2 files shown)


src/__tests__/unit/controllers/universal-router/test.tsx
```diff
@@ -126,7 +126,7 @@ describe('UniversalRouter', () => {
         .mockImplementation(() => mockHistory);
     });
 
-    it('should not respond to history changes', () => {
```
albertogasparin, line 129: This test was passing only because the history being called was legacy history. Listening to memory router on SSR is fine, there is no danger in that

src/controllers/universal-router/index.tsx
```diff
@@ -10,17 +9,13 @@ import { BrowserHistory } from 'src/common/types';
...
-  location ? createMemoryHistory({ initialEntries: [location] }) : history;
-
 /**
  * Default prop provider for the RouterContainer.
  *
  */
 export class UniversalRouter extends Component<UniversalRouterProps> {
   static defaultProps = {
     isGlobal: true,
-    history: DEFAULT_HISTORY,
```
albertogasparin, line 23: Bye bye `DEFAULT_HISTORY`

src/controllers/universal-router/index.tsx
```diff
@@ -88,7 +88,7 @@ export class UniversalRouter extends Component<UniversalRouterProps> {
     return (
       <UniversalRouterContainer
         routes={routes}
-        history={getInferredHistory(history, location)}
```
albertogasparin, line 91: This was creating a new history on every re-render. Better store it in the class so we avoid such risk

src/controllers/universal-router/index.tsx
```diff
@@ -56,12 +50,17 @@ export class UniversalRouter extends Component<UniversalRouterProps> {
...
+  history: BrowserHistory;
+
+  constructor(props: UniversalRouterProps) {
+    super(props);
+    const initialEntries = props.location ? [props.location] : [];
+    this.history = props.history || createMemoryHistory({ initialEntries });
+  }
 
   componentDidMount() {
-    if (!isServerEnvironment()) {
```
albertogasparin, line 61: No need to check a `componentDidMount` is not called on server

PR #31, opened 2020-07-06
Rewrite useResource to support custom context

albertogasparin:
Some of this might be unnecessary, but I think we should clean up a couple of things now, before adoption grows even more.

The core piece of this PR is the rewrite of `useResource` hook to support custom router context. Basically we allow to provide `route`, `match` and `query` to manipulate the resource with another generated key instead of based on current route location. To be honest, I find a bit annoying that we have to pass in all that data to generate the key, but don't see any other option (suggestion welcome though).  
Took also the chance to properly handle route changes: by generating a sweet-state hook on the fly we can subscribe to route changes but thanks to the selector the resource hook will trigger a re-render only if the key changes. 

```js
const MyComponent = () => {
  const { data, loading } = useResource(issueResource, {
    routerContext: createRouterContext(issueRoute, { issueKey: 'BLA-1' })
  });
  // ...
}
```

Related change is removing `location` from the passed in args to resources: it contains duplicated information and we can export a `generateLocation` ([like this](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/generatePath.js)) to easily build that if needed. This is to reduce the amount of input required when consumers want to provide a custom router context.

Another thing that I always found annoying is the array destructuring, so it's time to drop it. This will require consumers changes during the bump. We could still support both beading a `Symbol.iterator` method so that we create an array like object, however I wonder if we can make types happy and if it worth the additional complexity.

```js
const output = { ...slice, update, refresh };

return {
    ...output,

    // support the deprecated [{ data }] style
    0: output,
    length: 1,
    [Symbol.iterator]() {
      return {
        next: () => ({ value: output, done: false }),
      };
    },
};
```

TODO:
- [x] add tests
- [x] rewrite subscriber
- [x] fix flow types
- [x] export generatePath util


rohan-deshpande:
Nice! Can you give an example or have a test for how this will get used?

src/controllers/hooks/resource-store/index.tsx
```diff
@@ -16,45 +19,58 @@ type UseResourceHookResponse<
...
+      actions.updateResourceState(resource.type, key, resource.maxAge, updater);
     },
-  ];
+    [resource, key, actions]
+  );
+
+  // we keep route context bound to key, so route context won't refresh
+  // unless key changes. This allows refresh to be called on effect cleanup
+  // or asynchronously, when route context might already have changed
+  const routerContext = useMemo(
```
jackbrown, line 66: do you think we could get an advantage here by using something like `useMemoOne` which guarantees the value won't be updated until the array value changes rather than leaving this open to being cleaned up by React? Same for the `useCallback`.

albertogasparin: That's not enough. Take for instance `/projects/:projectKey/board/:boardId`: the match object will be `{ projectKey, boardId }` but if with `getKey: ({ match }) => match.projectKey` we don't care about `boardId` value.
So on `boardId` change we don't want to generate new values for `update` and `refresh` because the "slice" they are bound to is still the same.

albertogasparin:
Given `routerContext` is a quite complex object, I've exported a new utility called `createRouterContext` that allows to generate the proper object with match, query and route. 

I've also renamed current `getRouteContext` to `findRouterContext` as was causing confusion: this one needs a list of routes and location to find out which one is the current, hence I think "find" is a better verb. But happy to gain more feedback on naming and APIs.

src/common/utils/create-router-context/index.ts
```diff
@@ -0,0 +1,19 @@
+import { DEFAULT_MATCH } from '../../constants';
+import { MatchParams, Query, Route } from '../../types';
+import matchRoute from '../match-route';
+import generatePath from '../generate-path';
+
+export const createRouterContext = (
+  route: Route,
+  params: MatchParams = {},
```
pksjce, line 8: So the gist is I'd be able to pass in a custom key here to store different sets of data for the same resource + route?

albertogasparin: Well, we still go through `getKey` in the resource, as the url parameters are also used by `getData`. If we just specify the key, we don't know how to call `refresh` for instance, as `getData` needs the same router context.
So instead we require all the bits that might be needed by  `getKey` and `getData` so we can safely operate on the resource without being worried

jackbrown:
@albertogasparin this looks good to me. If you update the branch I'll ✅

src/common/utils/router-context/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  const matchedRoute = matchRoute([route], pathname, query);
+
+  return {
+    route,
+    match: matchedRoute ? matchedRoute.match : DEFAULT_MATCH,
+    query,
+  };
+};
+
+export const findRouterContext = (
```
albertogasparin, line 30: This is `getRouteContext` renamed

docs/api/hooks.md
```diff
@@ -31,6 +31,7 @@ export const Feed = () => {
 };
 ```
 
+It also acceps some options as second argument to customise the behaviour, like `routerContext`.
```
alifa20, line 34: It also accepts some

src/index.ts
```diff
@@ -19,7 +19,8 @@ export {
   matchRoute,
   generatePath,
   createLegacyHistory,
-  getRouteContext,
+  createRouterContext,
+  findRouterContext,
```
jackbrown, line 23: can we keep the `getRouteContext` export and add a console.warn suggesting that consumers swap to `findRouterContext`? I think it's better to keep the API backward compatible as much as possible.

src/index.ts
```diff
@@ -30,7 +31,7 @@ export type {
   MatchedRoute,
   RouteResource,
   RouteContext,
-  RouterStoreContext,
+  RouterContext,
```
jackbrown, line 34: same comment here. can we export both for now?

src/index.ts
albertogasparin: I'm happy to write a breaking api / migration doc in the changelog, but given we are still v0.x, maintaining it right now would mean it will never be changed. Is it used in Bitbucket? Jira does not have usages

src/index.ts
albertogasparin: If we don't break the API now, how do we plan migration late on? This is 0.x and I'd like having several APIs cleaned up before v1. I think doing bit by bit with a bunch of small releases is better than a big 1.0 break.

src/common/utils/router-context/index.ts
rohan-deshpande: y tho?

albertogasparin: Because it was wrong: we have both `RouterContext` and `RouteContext` as types. `RouterContext` now is just `route`, `match` and `query`, while `RouteContext` has also `location` and `action`. But `getRouteContext` never returned a valid `RouteContext` (as was missing `action`). It was returning an old `RouterContext` (with `location`), so it should have been named `getRouterContext` to be correct. So given the name should be fixed anyway, I'm making it more explicit given it accepts an array of routes as input.

src/index.ts
jackbrown: okay no problem. Happy with this change :)

PR #32, opened 2020-07-10
add router hooks

jackbrown:
This is to give us good compatibility with existing routers and make adoption easier :)

### DIFF
--- a/src/controllers/hooks/router-store/index.tsx
+++ b/src/controllers/hooks/router-store/index.tsx
@@ -1,5 +1,13 @@
 import { useRouterStore, useRouterStoreStatic } from '../../router-store';
 import { RouterActionsType, RouterState } from '../../router-store/types';
+import {
+  BrowserHistory,
+  MatchParams,
+  Match,
+  Route,
+  Location,
+} from '../../../common/types';
+import matchPath from '../../../common/utils/match-route/matchPath';
 
 /**
  * Utility hook for accessing the public router store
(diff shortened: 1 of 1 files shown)


src/controllers/hooks/router-store/index.tsx
```diff
@@ -18,3 +26,32 @@ export const useRouterActions = (): RouterActionsType => {
...
   return allActions;
 };
+
+export const useLocation = (): Location => {
+  const [routerState] = useRouterStore();
+
+  return routerState.location;
+};
+
+export const useHistory = (): BrowserHistory => {
```
albertogasparin, line 36: We always said that passing history around was a bad idea. Are we changing idea now?

src/controllers/hooks/router-store/index.tsx
```diff
@@ -18,3 +26,32 @@ export const useRouterActions = (): RouterActionsType => {
...
+  const [routerState] = useRouterStore();
+
+  return routerState.history;
+};
+
+export const useParams = (): MatchParams => {
```
albertogasparin, line 42: If we expose something like this, let's make it more powerful: allow a white list of params to look after, so consumers re-render only when one of those params change. Also, should we expose some actions with it to update them and change the url?

Moreover, I would do the same thing for `useQueryParams()` so we can get that for query parameters. And would be cool to have actions to add/remove/update the query string too

src/controllers/hooks/router-store/index.tsx
```diff
@@ -18,3 +26,32 @@ export const useRouterActions = (): RouterActionsType => {
...
+  const [routerState] = useRouterStore();
+
+  return routerState.match.params;
+};
+
+export const useRouteMatch = (route?: string | Route): Match => {
```
albertogasparin, line 48: I'm not entirely sure we should expose this. Do we have an actual use case?

src/controllers/hooks/router-store/index.tsx
rohan-deshpande: Yeah I agree here, I think we should be careful about exposing it

src/controllers/hooks/router-store/index.tsx
rohan-deshpande: What is the difference here between this and `useRouter`? Can't we get the current match from there?

jackbrown: there's no difference really, this is more of a compatibility thing. I'll remove this one and `useHistory` as we don't really need to support them.

PR #33, opened 2020-07-20
fix type exports for older TS versions

jackbrown:
This fixes an issue for consumers who are using older versions of TS that aren't compatible with `export type {...}` syntax

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -22,7 +22,7 @@ export {
   getRouteContext,
 } from './common/utils';
 
-export type {
+export {
   Location,
   Route,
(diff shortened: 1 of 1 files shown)


jackbrown:
after a chat with @MonicaOlejniczak we might need to put a bit more time toward this issue so that we don't end up breaking integration with CRA

rohan-deshpande:
Yep, I do recall some issues with CRA when not doing things this way, what was the reason for needing to support older versions of TS?

jackbrown:
We're going about this in a different manner

PR #36, opened 2020-07-21
Add utils entrypoint and new InvariantRoute type that doesn't have React dependency

### LINKED ISSUE #34: Support entrypoint that doesn't require React
We want to import the `matchRoute` method from this package on our Node.js service and we don't want to have to include React as a peer dependency. 

This can be supported by introducing a new entrypoint into the package.

cc @albertogasparin

JakeLane:
Adding a new utils entrypoint which is intended to allow importing of this package without adding react as a dependency.

This involved adding a few new types for routes, `InvariantRoute` and friends. This new type is the base type for `Route` and allows the type to exist without React (required by the `component` field on `Route`).

Resolves #34


utils/package.json
```diff
@@ -0,0 +1,5 @@
+{
```
JakeLane, line 1: Borrowing this pattern from [react-loosely-lazy](https://github.com/atlassian-labs/react-loosely-lazy/blob/master/babel-plugin/package.json), thanks @MonicaOlejniczak!

src/common/utils/match-route/index.ts
```diff
@@ -118,3 +126,15 @@ export default (
 
   return matchedRoute;
 };
+
+export const matchInvariantRoute = (
```
JakeLane, line 130: Two new overloads here

src/index.js.flow
```diff
@@ -1,297 +1,80 @@
...
+  RouterActionReplace,
+  RouterActionsType,
+  RouteResource,
+  RouteResourceError,
+  RouteResourceResponse,
+  RouterProps,
+  RouterStoreContext,
+  RouterSubscriberProps,
+  Routes,
+} from './types.js.flow';
```
MonicaOlejniczak, line 77: You shouldn't need to specify the full path here and in the other files. `./types` should be sufficient since flow will match the correct file for you. I would also make sure that you integrate this in Jira and make sure the types work as expected.

src/common/types.ts
```diff
@@ -110,16 +115,15 @@ export type RouterStoreContext = {
...
   exact?: boolean;
 
-  /** The component to render on match, typed explicitly */
-  component: ComponentType<RouteContext>;
-
-  /** If present, must return true to include the route. */
-  enabled?: () => boolean;
-
   /** Signals that this is a redirect route, we need to handle these in a special way. */
   isRedirect?: boolean;
```
jackbrown, line 128: nit; this isn't being used any more. Would you mind removing it?

JakeLane: Done 👍

src/index.js.flow
JakeLane: I've corrected this and tested it in Jira frontend without issue

src/utils.ts
```diff
@@ -0,0 +1,7 @@
+export {
+  createLegacyHistory,
+  generatePath,
+  getRouteContext,
```
jackbrown, line 4: this will need updating as the function was renamed in #31

PR #38, opened 2020-08-19
FIX: CRA/AF TypeScript incompatibility

flexdinesh:
Fixes incompat issues with TS versions < 3.9 —

- `export type` syntax was introduced in 3.8 and it prevented users with lower TS versions from consuming the package
- CRA builds crash because of undefined type imports

### DIFF
--- a/src/common/types.ts
+++ b/src/common/types.ts
@@ -2,9 +2,9 @@
 
 import { ComponentType, ReactNode } from 'react';
 
-import { History, Location as LocationShape } from 'history';
+import { History, Location as HistoryLocationShape } from 'history';
 
-export { LocationShape };
+export type LocationShape = HistoryLocationShape;
 
 export type Href = string;
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,5 @@
+export { createBrowserHistory, createMemoryHistory } from 'history';
+
 export {
   Router,
   MemoryRouter,
@@ -14,7 +16,9 @@ export {
   createResource,
   useRouterActions,
 } from './controllers';
+
 export { RouteComponent, Link, noopRouterDecorator } from './ui';
+
 export {
   matchRoute,
   generatePath,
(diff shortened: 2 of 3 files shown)


src/common/types.ts
```diff
@@ -2,9 +2,9 @@
 
 import { ComponentType, ReactNode } from 'react';
 
-import { History, Location as LocationShape } from 'history';
+import { History, Location as HistoryLocationShape } from 'history';
 
-export { LocationShape };
+export type LocationShape = HistoryLocationShape;
```
albertogasparin, line 7: I assume export type here is fine because we export `LocationShape` (without `type`) from index, right?

albertogasparin (approved):

jackbrown (approved):

rohan-deshpande:
Hey a bit confused here, are we saying CRA builds crash?

src/index.ts
```diff
@@ -23,7 +27,7 @@ export {
   findRouterContext,
 } from './common/utils';
 
-export type {
+export {
```
jackbrown, line 30: @rohan-deshpande the reason we don't need `export type` here, is because in `src/common/types` we are exporting `LocationShape` as a type now. 

Previously TS was unable to determine if `LocationShape` was a type or an actual JS property so the `export type` here was covering up the actual issue by forcing TS to treat it as a type.

src/common/types.ts
flexdinesh: This change fixed the `Attempted import error` with CRA. For some reason, tsc compiled output was not stripping out the `LocationShape` type from `types.js` after compilation. This change got it working right.

flexdinesh:
@rohan-deshpande Yup. Two things failed in CRA builds — 

1. `export type` syntax 'cos CRA has TS 3.7.2 and `export type` syntax was introduced in TS `3.8`.
2. `Attempted import error` happened because TS in CRA was unable some types possibly because of the way `LocationShape` was resolved.

PR #41, opened 2020-08-26
Local dev loop — webpack-dev-server with examples

flexdinesh:
I have setup a local dev server with examples to help new starters quickly build a mental model of the code and architecture. It also enables debugging with react dev tools.


rohan-deshpande:
Nice! maybe remove the codesandbox examples from the docs and include some way to run these for consumers?

flexdinesh:
@rohan-deshpande I think codesandbox examples add value for first-time users and visiters who want to take a look at how the API works without having to check out the repo.

Since the dev server is on `start` script and doesn't need any additional config to get started we can leave it be and maybe revisit when we want to put out an elaborate **Get Started** guide. Thoughts?

examples/BasicRoutingWithResources/home.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Link, createResource, useResource } from 'react-resource-router';
+
+export const homeResource = createResource({
+  type: 'home',
+  getKey: () => 'data',
+  maxAge: 0,
+  getData: async () => {
+    let result;
+    // eslint-disable-next-line
+    try {
```
albertogasparin, line 11: Having this `try/catch` here is actually a bad pattern, as if you forget to re-throw you want get the error status. I suggest we remove it

jackbrown (approved):
Looks good. Thanks for adding these!

albertogasparin (commented):
Can we grab the version of web pack config that I have in `react-sweet-state`? So we can run multiple examples without being limited only one at time?

rohan-deshpande:
> @rohan-deshpande I think codesandbox examples add value for first-time users and visiters who want to take a look at how the API works without having to check out the repo.

My only concern here is single source of truth insofar as if some API changes occur, we would now need to make changes in two places, both the local dev loop and the codesandbox. If stuff is simple enough for this to probably never be a problem then all good. 

> Since the dev server is on `start` script and doesn't need any additional config to get started we can leave it be and maybe revisit when we want to put out an elaborate **Get Started** guide. Thoughts?

Right but I think it's still a good idea to let users know we do have a local dev loop and this is how you use it, can just be a one liner in the docs.

examples/BasicRoutingWithResources/home.tsx
flexdinesh: I was wondering the same. Removed it now.

flexdinesh:
@rohan-deshpande It makes more sense now. We recently changed an API and updated it in the docs and left out codesandbox. I removed the codesandbox links from README.md and added instructions to run the dev server in its place.

flexdinesh:
@albertogasparin I ported the dev server config to auto generate examples and it looks lit now. Pushed in changes.

webpack.config.js
```diff
@@ -0,0 +1,63 @@
...
+  entry: generateExampleEntries(),
+
+  devServer: {
+    contentBase: path.resolve(__dirname, 'examples'),
+    publicPath: '/',
+  },
+
+  output: {
+    filename: '[name].js',
+    path: path.resolve(__dirname, 'dist'),
```
jackbrown, line 39: could you add this directory to `.gitignore`?

flexdinesh: `webpack-dev-server` uses memory instead of file system for serving the files. So no files or directories will be created in the disk. But I've added `dist` to `.gitignore` anyway. Pushing in the code shortly.

jackbrown (approved):

albertogasparin (approved):

PR #42, opened 2020-09-04
Add new hooks - useQueryParam and usePathParam

flexdinesh:
Added a new hooks `useQueryParam` and `usePathParam` to access and update query and path params respectively in current route. Changes include —

- Two new hooks
- Example to demonstrate hooks usage
- Unit tests
- API docs

```js
import { useQueryParam } from 'react-resource-router';

// Current route in address bar — /home/projects?foo=bar

export const MyComponent = () => {
  const [foo, setFoo] = useQueryParam('foo');
  // => foo will have the value 'bar'

  const clickHandler = () => {
    setFoo('baz');
    // => Will update current route to /home/projects?foo=baz
  };

  return (
    <div>
      <p>Hello World!</p>
      <button onClick={clickHandler}>Update param</button>
    </div>
  );
};
```

```js
import { usePathParam } from 'react-resource-router';

// path — /projects/:projectId/board/:boardId

// Current route in address bar — /projects/123/board/456?foo=bar

export const MyComponent = () => {
  const [projectId, setProjectId] = usePathParam('foo');
  // => projectId will have the value '123'

  const clickHandler = () => {
    setProjectId('xyz');
    // => Will update current route to /projects/xyz/board/456?foo=bar
  };

  return (
    <div>
      <p>Hello World!</p>
      <button onClick={clickHandler}>Update param</button>
    </div>
  );
};
```

### DIFF
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -61,3 +61,30 @@ export const RouterActionsHookExample = () => {
   return <MyComponent push={push} />;
 };
 ```
+
+## useQueryParam
+
+You can use the `useQueryParam` hook to access the query params in current route.
+
+```js
+import { useQueryParam } from 'react-resource-router';
+
+// Current route in address bar — /home/projects?foo=bar
+
+export const MyComponent = () => {
+  const [foo, setFoo] = useQueryParam('foo');
+  // => foo will have the value 'bar'
+
+  const clickHandler = () => {
+    setFoo('baz');
+    // => Will update current route to /home/projects?foo=baz
+  };
+
+  return (
+    <div>
+      <p>Hello World!</p>
+      <button onClick={clickHandler}>Update param</button>
+    </div>
+  );
+};
+```
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -64,7 +64,7 @@ export const RouterActionsHookExample = () => {
 
 ## useQueryParam
 
-You can use the `useQueryParam` hook to access the query params in current route.
+You can use the `useQueryParam` hook to access the query params in current route. Pass in `undefined` to remove a query param from the url.
 
 ```js
 import { useQueryParam } from 'react-resource-router';
@@ -78,6 +78,35 @@ export const MyComponent = () => {
   const clickHandler = () => {
     setFoo('baz');
     // => Will update current route to /home/projects?foo=baz
+  };
+
+  return (
+    <div>
+      <p>Hello World!</p>
+      <button onClick={clickHandler}>Update param</button>
+    </div>
+  );
+};
+```
+
+## usePathParam
+
+You can use the `usePathParam` hook to access the path params in current route. Pass in `undefined` to remove an [optional param](https://github.com/pillarjs/path-to-regexp#optional) from the url.
+
+```js
+import { usePathParam } from 'react-resource-router';
+
+// path — /projects/:projectId/board/:boardId
+
+// Current route in address bar — /projects/123/board/456?foo=bar
+
+export const MyComponent = () => {
+  const [projectId, setProjectId] = usePathParam('projectId');
+  // => projectId will have the value '123'
+
+  const clickHandler = () => {
+    setProjectId('222');
+    // => Will update current route to /projects/222/board/456?foo=bar
   };
 
   return (
--- a/examples/hooks/index.tsx
+++ b/examples/hooks/index.tsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import {
+  Router,
+  RouteComponent,
+  createBrowserHistory,
+} from 'react-resource-router';
+
+import Home from './home';
+import QueryParamExample from './use-query-param';
+
+const myHistory = createBrowserHistory();
+
+const baseURL = '/hooks';
+
+const appRoutes = [
+  {
+    name: 'home',
+    path: `${baseURL}`,
+    exact: true,
+    component: Home,
+    navigation: null,
+  },
+  {
+    name: 'query-param',
+    path: `${baseURL}/query-param`,
+    exact: true,
+    component: QueryParamExample,
+    navigation: null,
+  },
+];
+
+const App = () => {
+  return (
+    <Router routes={appRoutes} history={myHistory}>
+      <RouteComponent />
+    </Router>
+  );
+};
+
+ReactDOM.render(<App />, document.getElementById('root'));
--- a/examples/hooks/index.tsx
+++ b/examples/hooks/index.tsx
@@ -9,6 +9,7 @@ import {
 
 import Home from './home';
 import QueryParamExample from './use-query-param';
+import PathParamExample from './use-path-param';
 
 const myHistory = createBrowserHistory();
 
@@ -27,6 +28,13 @@ const appRoutes = [
     path: `${baseURL}/query-param`,
     exact: true,
     component: QueryParamExample,
+    navigation: null,
+  },
+  {
+    name: 'path-param',
+    path: `${baseURL}/path-param/:foo/:bar`,
+    exact: true,
+    component: PathParamExample,
     navigation: null,
   },
 ];
--- a/src/__tests__/unit/controllers/hooks/router-store/test.tsx
+++ b/src/__tests__/unit/controllers/hooks/router-store/test.tsx
@@ -1,0 +1,164 @@
+import React from 'react';
+
+import { mount } from 'enzyme';
+import * as historyHelper from 'history';
+import { defaultRegistry } from 'react-sweet-state';
+import { act } from 'react-dom/test-utils';
+
+import { MemoryRouter } from '../../../../../controllers/memory-router';
+import { useQueryParam } from '../../../../../controllers/hooks/router-store';
+
+const mockLocation = {
+  pathname: '/pathname',
+  search: '?foo=bar&hello=world',
+  hash: '#hash',
+};
+
+const mockHistory = {
+  push: jest.fn(),
+  replace: jest.fn(),
+  goBack: jest.fn(),
+  goForward: jest.fn(),
+  registerBlock: jest.fn(),
+  listen: jest.fn(),
+  createHref: jest.fn(),
+  location: mockLocation,
+  block: jest.fn(),
+};
+
+const mockRoutes = [
+  {
+    path: '/pathname',
+    component: () => <div>path</div>,
+    name: '',
+  },
+  {
+    path: '/blah',
+    component: () => <div>path</div>,
+    name: '',
+  },
+];
+
+const nextTick = () => new Promise(resolve => setTimeout(resolve));
+
+const MockComponent = ({ children, ...rest }: any) => {
+  return children(rest);
+};
+
+describe('useQueryParam', () => {
+  const { location } = window;
+
+  beforeAll(() => {
+    delete window.location;
+    // @ts-ignore
+    window.location = {};
+    Object.defineProperties(window.location, {
+      href: { value: location.href },
+      assign: { value: jest.fn() },
+      replace: { value: jest.fn() },
+    });
+  });
+
+  beforeEach(() => {
+    jest
+      .spyOn(historyHelper, 'createMemoryHistory')
+      // @ts-ignore
+      .mockImplementation(() => mockHistory);
+  });
+
+  afterEach(() => {
+    defaultRegistry.stores.clear();
+    jest.resetAllMocks();
+    jest.restoreAllMocks();
+  });
+
+  afterAll(() => {
+    window.location = location;
+  });
+
+  it('should return the right param value', () => {
+    mount(
+      <MemoryRouter routes={mockRoutes}>
+        <MockComponent>
+          {() => {
+            const [param] = useQueryParam('foo');
+            expect(param).toEqual('bar');
+
+            return <div>I am a subscriber</div>;
+          }}
+        </MockComponent>
+      </MemoryRouter>
+    );
+  });
+
+  it('should return undefined for non-existent params', () => {
+    mount(
+      <MemoryRouter routes={mockRoutes}>
+        <MockComponent>
+          {() => {
+            const [param] = useQueryParam('iamnotaqueryparam');
+            expect(param).toEqual(undefined);
+
+            return <div>I am a subscriber</div>;
+          }}
+        </MockComponent>
+      </MemoryRouter>
+    );
+  });
+
+  it('should update URL with new param value', async () => {
+    const mockPath = mockLocation.pathname;
+    let qpVal: string | undefined, qpUpdateFn: (qp: string) => void;
+
+    mount(
+      <MemoryRouter routes={mockRoutes}>
+        <MockComponent>
+          {() => {
+            const [param, setParam] = useQueryParam('foo');
+            qpVal = param;
+            qpUpdateFn = setParam;
+
+            return <div>I am a subscriber</div>;
+          }}
+        </MockComponent>
+      </MemoryRouter>
+    );
+
+    expect(qpVal).toEqual('bar');
+
+    act(() => qpUpdateFn('newVal'));
+
+    await nextTick();
+
+    expect(mockHistory.push).toBeCalledWith(
+      `${mockPath}?hello=world&foo=newVal`
+    );
+  });
+
+  it('should remove param from URL when set to null', async () => {
+    const mockPath = mockLocation.pathname;
+    let qpVal: string | undefined, qpUpdateFn: (qp: string | null) => void;
+
+    mount(
+      <MemoryRouter routes={mockRoutes}>
+        <MockComponent>
+          {() => {
+            const [param, setParam] = useQueryParam('foo');
+            qpVal = param;
+            qpUpdateFn = setParam;
+
+            return <div>I am a subscriber</div>;
+          }}
+        </MockComponent>
+      </MemoryRouter>
+    );
+
+    expect(qpVal).toEqual('bar');
+
+    act(() => qpUpdateFn(null));
+
+    await nextTick();
+
+    expect(mockHistory.push).toBeCalledWith(`${mockPath}?hello=world`);
+  });
+});
--- a/src/__tests__/unit/controllers/hooks/router-store/test.tsx
+++ b/src/__tests__/unit/controllers/hooks/router-store/test.tsx
@@ -5,30 +5,22 @@ import * as historyHelper from 'history';
 import { defaultRegistry } from 'react-sweet-state';
 import { act } from 'react-dom/test-utils';
 
-import { MemoryRouter } from '../../../../../controllers/memory-router';
-import { useQueryParam } from '../../../../../controllers/hooks/router-store';
+import { Router } from '../../../../../controllers';
+import {
+  useQueryParam,
+  usePathParam,
+} from '../../../../../controllers/hooks/router-store';
+import { getRouterStore } from '../../../../../controllers/router-store';
 
 const mockLocation = {
-  pathname: '/pathname',
-  search: '?foo=bar&hello=world',
+  pathname: '/projects/123/board/456',
+  search: '?foo=hello&bar=world',
   hash: '#hash',
-};
-
-const mockHistory = {
-  push: jest.fn(),
-  replace: jest.fn(),
-  goBack: jest.fn(),
-  goForward: jest.fn(),
-  registerBlock: jest.fn(),
-  listen: jest.fn(),
-  createHref: jest.fn(),
-  location: mockLocation,
-  block: jest.fn(),
 };
 
 const mockRoutes = [
   {
-    path: '/pathname',
+    path: '/projects/:projectId/board/:boardId',
     component: () => <div>path</div>,
     name: '',
   },
@@ -39,6 +31,14 @@ const mockRoutes = [
   },
 ];
 
+const historyBuildOptions = {
+  initialEntries: [
+    `${mockLocation.pathname}${mockLocation.search}${mockLocation.hash}`,
+  ],
+};
+
+let history = historyHelper.createMemoryHistory(historyBuildOptions);
+let historyPushSpy = jest.spyOn(history, 'push');
 const nextTick = () => new Promise(resolve => setTimeout(resolve));
 
 const MockComponent = ({ children, ...rest }: any) => {
@@ -46,24 +46,9 @@ const MockComponent = ({ children, ...rest }: any) => {
 };
 
 describe('useQueryParam', () => {
-  const { location } = window;
-
-  beforeAll(() => {
-    delete window.location;
-    // @ts-ignore
-    window.location = {};
-    Object.defineProperties(window.location, {
-      href: { value: location.href },
-      assign: { value: jest.fn() },
-      replace: { value: jest.fn() },
-    });
-  });
-
   beforeEach(() => {
-    jest
-      .spyOn(historyHelper, 'createMemoryHistory')
-      // @ts-ignore
-      .mockImplementation(() => mockHistory);
+    history = historyHelper.createMemoryHistory(historyBuildOptions);
+    historyPushSpy = jest.spyOn(history, 'push');
   });
 
   afterEach(() => {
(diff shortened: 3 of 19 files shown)


docs/api/hooks.md
```diff
@@ -61,3 +61,23 @@ export const RouterActionsHookExample = () => {
...
+```js
+import { useQueryParam } from 'react-resource-router';
+
+// Current route in address bar — /home/projects?foo=bar
+
+export const MyComponent = () => {
+  const [foo, setFoo] = useQueryParam('foo');
+  // => foo will have the value 'bar'
+
+  setFoo('baz');
```
theKashey, line 78: 🤯

flexdinesh: @theKashey is that a "that's awesome" 🤯 or "what the hell are you doing here?" 🤯?

flexdinesh: Oh yeah, the example is non contextual. I'll update it to invoke the setter on a button event.

package.json
```diff
@@ -68,6 +68,7 @@
     "lodash.debounce": "^4.0.8",
     "lodash.noop": "^3.0.1",
     "path-to-regexp": "^1.7.0",
+    "query-string": "^6.13.1",
```
albertogasparin, line 71: No need for this. `url-parse` already exposes a version and we should use that instead of adding another dependency

src/__tests__/unit/controllers/hooks/router-store/test.tsx
```diff
@@ -0,0 +1,161 @@
...
+    );
+  });
+
+  it('should remove param from URL when set to null', async () => {
+    const mockPath = mockLocation.pathname;
+    let qpVal: string | undefined, qpUpdateFn: (qp: string | null) => void;
+
+    mount(
+      <MemoryRouter routes={mockRoutes}>
+        <RouterSubscriber>
```
albertogasparin, line 141: Why do we need `RouterSubscriber` (here and in the other tests)?

src/controllers/hooks/router-store/index.tsx
```diff
@@ -18,3 +20,36 @@ export const useRouterActions = (): RouterActionsType => {
...
   return allActions;
 };
+
+/**
+ * Utility hook for accessing URL query params
+ */
+export const useQueryParam = (
+  paramKey: string
+): [string | undefined, (qp: string | null) => void] => {
+  const [entireState, routerActions] = useRouterStore();
```
albertogasparin, line 30: I think this will always re-render. We need to create a dedicated hook with a selector to avoid that.

webpack.config.js
```diff
@@ -32,6 +32,9 @@ module.exports = {
   devServer: {
     contentBase: path.resolve(__dirname, 'examples'),
     publicPath: '/',
+    // historyApiFallback: {
+    //   rewrites: [{ from: /\/hooks\/*/, to: '/hooks/index.html' }],
+    // },
```
albertogasparin, line 37: 🤔

albertogasparin (changes requested):

package.json
flexdinesh: Thanks for the heads up. I've removed `query-string` dep.

src/__tests__/unit/controllers/hooks/router-store/test.tsx
flexdinesh: We don't. I took inspiration for the test approach from other tests and ported that as copypasta. I have removed it now and tests behave the same. 👍

webpack.config.js
flexdinesh: When reloading pages, we get `404` from webpack-dev-server 'cos those URLs (`/hooks/use-query-param`) don't exist and it was a bit of an inconvenience. This hack let's us use the reload workflow but prevents `contentBase` serving the root examples we generate. I left the comment here so I can uncomment it just for hooks development.

src/controllers/hooks/router-store/index.tsx
flexdinesh: Nice catch. I re worked on the logic using a selector. Force pushed the code.

src/controllers/hooks/router-store/index.tsx
```diff
@@ -18,3 +31,43 @@ export const useRouterActions = (): RouterActionsType => {
...
+  paramKey: string
+): [string | undefined, (qp: string | null) => void] => {
+  const [paramVal, routerActions] = useParam({ paramKey });
+
+  const setQueryParam = React.useCallback(
+    (newValue: string | undefined | null) => {
+      const {
+        location: { pathname },
+        query,
+      } = getRouterState();
```
flexdinesh, line 57: @albertogasparin Is it a good idea to access the store with `getRouterState` here? We need the `query` object for non-reactive read only purpose here. If there's a better approach lemme know pl.

webpack.config.js
theKashey: It can be just `historyApiFallback: true`, to fallback to `/index.html` and let "router" handle the rest in a "SPA mode"

src/controllers/hooks/router-store/index.tsx
albertogasparin: No, I don't think it is the best approach. I think it could be much better to move all this query (and path) manipulations to a sweets-state action, so we don't have to make the entire store available in the hook (`getRouterState` is an anti pattern)

We can either create specific `pushQuery`/`pushPath` actions to manipulate the url and call the generic `push`, or we can extend `push` directly and allow it to be called with an object. Something like:

```js
push({ param: { key: value }, query: { key: value }  });
```

The good thing about the latter is that we can in future allow using push without requiring string manipulation from the consumer side, as the API could evolve to allow:

```js
push({ route, param: { key: value }, query: { key: value }  });
// also replace could allow such shape
replace({ route, param: { key: value }  });
```

The main thing that we should pay attention at is how do we communicate from the API when `param/query` keys are merged with existing ones or replaced. We might need a specific option to make it obvious.

What I mean is, assuming we have `param: { foo: 1, bar: 1 }`, how do we allow consumers to say with `push({ param: { foo: 2 } })` that they want as a result `param: { foo: 2, bar: 1 }` or just `param: { foo: 2 }`. 

Maybe something like `push({ param: { ... }, query: { ... }, merge: true })` ?
Or `partial: true`? Or `override: false`? Or ideas?

src/__tests__/unit/controllers/hooks/router-store/test.tsx
```diff
@@ -0,0 +1,164 @@
...
+    );
+
+    expect(qpVal).toEqual('bar');
+
+    act(() => qpUpdateFn(null));
+
+    await nextTick();
+
+    expect(mockHistory.push).toBeCalledWith(`${mockPath}?hello=world`);
+  });
```
albertogasparin, line 163: Remember to add a unit test to ensure that components listening to different query parameters are not re-rendered if their parameter does not change

src/controllers/hooks/router-store/index.tsx
```diff
@@ -18,3 +31,43 @@ export const useRouterActions = (): RouterActionsType => {
 
   return allActions;
 };
+
+const useParam = createHook<
```
albertogasparin, line 35: This should be extended to be used from `usePathParam` too. We could make the selector accepting `{ paramKey, source }` or something similar so we can specify if we want the path parameter or the query parameter and avoid having another hook.

albertogasparin (changes requested):

webpack.config.js
flexdinesh: `historyApiFallback: true` doesn't work when we generate and serve static files as a directory using `contentBase` option in `devServer`. It always gives 404. Haven't figured why it does that though.

src/controllers/hooks/router-store/index.tsx
flexdinesh: @albertogasparin I'm glad I pointed this out. I've now removed `getRouterState` and created two new actions `pushQueryParam` and `pushPathParam` that does the job for the hooks. 

Re: extending the push action to support query and path params as API
I looked into this a bit and figured there are a few moving parts that need to be considered before exposing it as an api (eg. absolute URLs and others) and it leads to more code paths. I think we could revisit this when we decide if we are going to expose push action to support a config param.

src/__tests__/unit/controllers/hooks/router-store/test.tsx
flexdinesh: Added a unit test to check re-renders of hooked components

src/controllers/hooks/router-store/index.tsx
flexdinesh: I created two new two creators instead — `createQueryParamHook` and `createPathParamHook`

src/controllers/router-store/index.tsx
```diff
@@ -200,6 +211,38 @@ const actions: AllRouterActions = {
...
+      history,
+      location,
+      route: { path: rawPath },
+      match: { params: existingPathParams },
+    } = getState();
+    const updatedPathParams = { ...existingPathParams, ...params };
+    const updatedPath = generatePathUsingPathParams(rawPath, updatedPathParams);
+    const updatedLocation = { ...location, pathname: updatedPath };
+
+    history.push(getRelativePath(updatedLocation) as any);
```
albertogasparin, line 244: Can you explain why we need to call `getRelativePath` ? `updatedLocation` is an object so it should do nothing with it 🤔

albertogasparin (changes requested):

src/controllers/router-store/index.tsx
flexdinesh: Our `history.push()` takes only string as param. `getRelativePath` util deconstructs the `Location` object and safely returns the url constructed in a string format.

src/controllers/router-store/index.tsx
```diff
@@ -200,6 +211,38 @@ const actions: AllRouterActions = {
...
+        (updatedQueryParams[key] === undefined ||
+          updatedQueryParams[key] === null) &&
+        delete updatedQueryParams[key]
+    );
+    const updatedPath = updateQueryParams(
+      location,
+      updatedQueryParams as Query
+    );
+
+    history.push(getRelativePath(updatedPath) as any);
```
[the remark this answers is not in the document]
flexdinesh: After the changes, this `getRelativePath` util is not needed here. I mage the return of `updateQueryParams` util to be string. Force pushed.

src/controllers/router-store/utils/path-utils.ts
```diff
@@ -28,3 +28,13 @@ export const getRelativePath = (path: Href | Location): string | Location => {
...
 };
+
+export const updateQueryParams = (
+  location: Location,
+  query: Query
+): string | Location => {
+  const url = new URL(location.pathname + location.search + location.hash);
+  const stringifiedQuery = qs.stringify(query);
+
+  return `${url.pathname}?${stringifiedQuery}${url.hash}`;
```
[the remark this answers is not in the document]
flexdinesh: This is nice. I've made changes and added some tests around it as well.

src/controllers/router-store/index.tsx
flexdinesh: We have a new util now `getURLFromLocation` 😄

src/controllers/hooks/router-store/index.tsx
```diff
@@ -18,3 +29,58 @@ export const useRouterActions = (): RouterActionsType => {
...
+
+  return [paramVal, setQueryParam];
+};
+
+/**
+ * Utility hook for accessing URL path params
+ */
+export const usePathParam = (
+  paramKey: string
+): [string | undefined, (qp: string) => void] => {
```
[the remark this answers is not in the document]
flexdinesh: yup. changed it to pp.

src/controllers/router-store/utils/path-utils.ts
```diff
@@ -28,3 +28,13 @@ export const getRelativePath = (path: Href | Location): string | Location => {
 
   return `${url.pathname}${url.query}${url.hash}`;
 };
+
+export const updateQueryParams = (
+  location: Location,
+  query: Query
+): string | Location => {
+  const url = new URL(location.pathname + location.search + location.hash);
```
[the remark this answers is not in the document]
flexdinesh: we don't. It's copy pasta from the above util and I though it's required at the moment. After some digging in it turns out we just need to return the string to `history.push()`

src/__tests__/unit/controllers/hooks/router-store/test.tsx
```diff
@@ -0,0 +1,347 @@
...
+    );
+  });
+
+  it('should return undefined for non-existent params', () => {
+    mount(
+      <MemoryRouter routes={mockRoutes}>
+        <MockComponent>
+          {() => {
+            const [param] = useQueryParam('iamnotaqueryparam');
+            expect(param).toEqual(undefined);
```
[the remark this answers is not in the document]
flexdinesh: I'll make the changes but curious to know if this feedback is more about consistency in test approach or if there are downsides to asserting inside components. I've seen other tests assert inside components and took inspiration from them.

examples/hooks/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  },
+  {
+    name: 'query-param',
+    path: `${baseURL}/query-param`,
+    exact: true,
+    component: QueryParamExample,
+    navigation: null,
+  },
+  {
+    name: 'query-param',
```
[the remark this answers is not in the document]
flexdinesh: changed

docs/api/hooks.md
```diff
@@ -61,3 +61,59 @@ export const RouterActionsHookExample = () => {
...
+// path — /projects/:projectId/board/:boardId
+
+// Current route in address bar — /projects/123/board/456?foo=bar
+
+export const MyComponent = () => {
+  const [projectId, setProjectId] = usePathParam('foo');
+  // => projectId will have the value '123'
+
+  const clickHandler = () => {
+    setProjectId('xyz');
```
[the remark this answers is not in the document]
flexdinesh: changed

docs/api/hooks.md
```diff
@@ -61,3 +61,59 @@ export const RouterActionsHookExample = () => {
...
+
+```js
+import { usePathParam } from 'react-resource-router';
+
+// path — /projects/:projectId/board/:boardId
+
+// Current route in address bar — /projects/123/board/456?foo=bar
+
+export const MyComponent = () => {
+  const [projectId, setProjectId] = usePathParam('foo');
```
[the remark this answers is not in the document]
flexdinesh: changed

src/controllers/router-store/utils/path-utils.ts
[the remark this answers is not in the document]
flexdinesh: force pushed

src/__tests__/unit/controllers/hooks/router-store/test.tsx
```diff
@@ -0,0 +1,347 @@
...
+];
+
+const nextTick = () => new Promise(resolve => setTimeout(resolve));
+
+const MockComponent = ({ children, ...rest }: any) => {
+  return children(rest);
+};
+
+describe('useQueryParam', () => {
+  const { location } = window;
```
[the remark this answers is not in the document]
flexdinesh: This is great. I've simplified the test with `createMemoryHistory` and `Router` component.

src/__tests__/unit/controllers/hooks/router-store/test.tsx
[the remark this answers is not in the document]
flexdinesh: Move the assertion outside the component

src/__tests__/unit/controllers/hooks/router-store/test.tsx
```diff
@@ -0,0 +1,306 @@
...
+      '<div id="foo-count">2</div>'
+    );
+    expect(wrapper.find(ComponentBar).html()).toEqual(
+      '<div id="bar-count">2</div>'
+    );
+
+    act(() => fooUpdateFn('newVal'));
+    await nextTick();
+
+    // URL is now — /projects/123/board/456?foo=newVal&bar=world
```
jackbrown, line 205: can we assert on this? rather than doing 2 different `expects` just assert on the whole URL?

src/__tests__/unit/controllers/hooks/router-store/test.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+            return <div>I am a subscriber</div>;
+          }}
+        </MockComponent>
+      </Router>
+    );
+
+    expect(qpVal).toEqual('123');
+
+    act(() => qpUpdateFn('newVal'));
```
jackbrown, line 298: what happens the update function is called with `undefined` or `null`? Can we add a test for that?

src/__tests__/unit/controllers/hooks/router-store/test.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+            return <div>I am a subscriber</div>;
+          }}
+        </MockComponent>
+      </Router>
+    );
+    expect(ppVal).toEqual('123');
+  });
+
+  it('should return undefined for non-existent params', () => {
```
jackbrown, line 262: could you add a test that updates more than 1 path param? i.e. `projectId` and `boardId` We should test that only the intended param was updated

src/__tests__/unit/controllers/hooks/router-store/test.tsx
```diff
@@ -0,0 +1,306 @@
...
+            expect(param).toEqual('hello');
+
+            return <div>I am a subscriber</div>;
+          }}
+        </MockComponent>
+      </Router>
+    );
+  });
+
+  it('should return undefined for non-existent params', () => {
```
jackbrown, line 75: can you add a test for adding a new param? So it should return `undefined` in the first place, and then after calling `setParam` it should update the URL

src/__tests__/unit/controllers/hooks/router-store/test.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+            return <div>I am a subscriber</div>;
+          }}
+        </MockComponent>
+      </Router>
+    );
+
+    expect(qpVal).toEqual('hello');
+
+    act(() => qpUpdateFn(null));
```
jackbrown, line 139: I might be overthinking this, but is removing the querystring parameter the desired behavior?

I put together a quick sandbox, using `qs.stringify` and `history.push`. In the example I call `qs.stringify` with `{ foo: null}` and the path is updated to `/home?foo=` which seems easier to reason about.

You can see the sandbox here https://codesandbox.io/s/react-resource-router-basic-routing-example-forked-pyg2m?file=/src/home/index.tsx

src/controllers/router-store/index.tsx
```diff
@@ -200,6 +212,38 @@ const actions: AllRouterActions = {
 
     return { query, route, match };
   },
+
+  pushQueryParam: params => ({ getState }) => {
+    const { query: existingQueryParams, history, location } = getState();
+    const updatedQueryParams = { ...existingQueryParams, ...params };
+    // remove null/undefined keys
```
jackbrown, line 219: I left a comment on the test, but we should discuss if this behavior is desired. Was there a request to remove params that have a `null` or `undefined` value?

src/controllers/router-store/utils/path-utils.ts
```diff
@@ -28,3 +28,14 @@ export const getRelativePath = (path: Href | Location): string | Location => {
...
 };
+
+export const updateQueryParams = (location: Location, query: Query): string => {
+  // @ts-ignore stringify accepts two params but it's type doesn't say so
+  const stringifiedQuery = qs.stringify(query, true);
+
+  return `${location.pathname}${stringifiedQuery}${location.hash}`;
+};
+
+export const getURLFromLocation = (location: Location): string => {
```
jackbrown, line 39: nit; can this be named `getRelativeURLFromLocation`?

webpack.config.js
jackbrown, line 37: do we still need this?

jackbrown (changes requested):
Looking good overall @flexdinesh. Let me know if you want to chat through any of the changes.

src/controllers/router-store/utils/path-utils.ts
flexdinesh: Yup

webpack.config.js
flexdinesh: Probably not. I'll remove it.

src/__tests__/unit/controllers/hooks/router-store/test.tsx
flexdinesh: Since we're using `createMemoryHistory()` we'll have to dig up the url from store instance. Should be able to add another assertion statement for the whole url.

flexdinesh: done

src/__tests__/unit/controllers/hooks/router-store/test.tsx
flexdinesh: We can't update the path to have null or undefined. Path compilation package we use throws an error in that case. And I think that's expected. Do you think we should filter out `null` or `undefined` path params before updating the path? @jackbrown

src/__tests__/unit/controllers/hooks/router-store/test.tsx
flexdinesh: `usePathParam('key')` hook takes only one param at a time. There's no API way of updating multiple path params at the same time. @jackbrown

src/__tests__/unit/controllers/hooks/router-store/test.tsx
flexdinesh: Good call. I've added a test to cover this case. Surprisingly it worked without any changes. Goes to show that we got the API design right.

src/__tests__/unit/controllers/hooks/router-store/test.tsx
flexdinesh: This is something I came up with during API design. The motivation behind the decision to be able to remove the param from the URL is —
1. To be able to differentiate between empty value and non-existent value - `/path?foo=` vs `/path` where foo is non-existent as opposed to "we don't know if it's non-existent or empty".
2. I created a forked example of the codesandbox you added - https://codesandbox.io/s/react-resource-router-query-string-exp-forked-dxfcs?file=/src/home/index.tsx. You can see the output URL of both the sandboxes is the same `/about?foo=` but the value we set in your sandbox is null and the value we set in my sandbox is an empty string `''`. With the API design we created, users can choose to either pass an empty string to just remove the value or pass undefined/null to remove the param altogether.

Thoughts?

albertogasparin: I like @flexdinesh approach better, otherwise consumers cannot remove query parameters (and sometimes in Jira they have to)

src/__tests__/unit/controllers/hooks/router-store/test.tsx
```diff
@@ -0,0 +1,342 @@
...
+
+    await nextTick();
+
+    expect(historyPushSpy).toBeCalledWith(`${mockPath}?bar=world#hash`);
+  });
+
+  it('should only re-render components hooked to a specific param', async () => {
+    let fooVal: string | undefined,
+      barVal: string | undefined,
```
albertogasparin, line 172: Same as other comment before, let's use a `let` per declaration

src/controllers/router-store/index.tsx
albertogasparin: Yes, we should allow consumers to remove query parameters. However we should pick one: I'm worried that by allowing both `null` or `undefined` we complicate our lives.

Like what is the outcome of `useQueryParam('foo')` or `usePathParam('foo')` when `foo` is not defined? Is that `null` or `undefined`? We should be clear about that, consistent and we should add a couple of tests around it ensuring that the values returned are correct.

src/controllers/hooks/router-store/index.tsx
```diff
@@ -18,3 +29,58 @@ export const useRouterActions = (): RouterActionsType => {
...
+/**
+ * Utility hook for accessing URL path params
+ */
+export const usePathParam = (
+  paramKey: string
+): [string | undefined, (pp: string) => void] => {
+  const [paramVal, routerActions] = createPathParamHook({ paramKey });
+
+  const setPathParam = React.useCallback(
+    (newValue: string) => {
```
albertogasparin, line 79: We should support also `undefined`, as path can have optional values

src/__tests__/unit/controllers/hooks/router-store/test.tsx
flexdinesh: Changed to use a `let` per declaration

src/__tests__/unit/controllers/hooks/router-store/test.tsx
flexdinesh: We're working on making `usePathParam` to accept `undefined` in the setter. @albertogasparin pointed out that we can have optional path params and we're accommodating that case to our API.

flexdinesh: @albertogasparin @jackbrown `usePathParam` setter now takes `undefined` that will remove optional params from the url and throw when undefined is passed for non-optional params. Added tests around these cases as well.

src/controllers/hooks/router-store/index.tsx
flexdinesh: https://github.com/atlassian-labs/react-resource-router/pull/42#discussion_r492471177

albertogasparin (approved):

albertogasparin (approved):

src/__tests__/unit/controllers/hooks/router-store/test.tsx
jackbrown: Ah i meant use 2 different hooks in a single test. It's no biggie though I'm happy to leave it for now

src/__tests__/unit/controllers/hooks/router-store/test.tsx
jackbrown: Cool that simplifies it. Nice one @flexdinesh !

src/controllers/router-store/index.tsx
jackbrown: I'm happy with that. Supporting `null` and `undefined` does seem to complicate it for no good reason.

jackbrown (approved):
This is looking great. Thanks for combining both hooks in 1 PR and @albertogasparin for the review!

albertogasparin (approved):

PR #43, opened 2020-09-16
feat: better typescript typings for resources

theKashey:
_Better_ typescript types:
- `createResource` to provide types for `useResource`.
- more `stateful` resource data -> `.data` can be safely accessed only if `loading=false` and `error` is not defined.

### DIFF
--- a/src/__tests__/unit/controllers/router-store/test.tsx
+++ b/src/__tests__/unit/controllers/router-store/test.tsx
@@ -474,15 +474,15 @@ describe('SPA Router store', () => {
 
       const ComponentB = () => (
         <ResourceSubscriber resource={resourceB}>
-          {({ data, loading, error }) => {
-            if (loading) {
-              return <div>{loading}</div>;
+          {props => {
+            if (props.loading) {
+              return <div>{props.loading}</div>;
             }
-            if (error) {
-              return <div>{error}</div>;
+            if (props.error) {
+              return <div>{props.error}</div>;
             }
 
-            return <div>{data.toString()}</div>;
+            return <div>{props.data.toString()}</div>;
           }}
         </ResourceSubscriber>
       );
--- a/src/__tests__/unit/controllers/router-store/test.tsx
+++ b/src/__tests__/unit/controllers/router-store/test.tsx
@@ -474,12 +474,15 @@ describe('SPA Router store', () => {
 
       const ComponentB = () => (
         <ResourceSubscriber resource={resourceB}>
-          {({ data, loading }) => {
+          {({ data, loading, error }) => {
             if (loading) {
               return <div>{loading}</div>;
             }
+            if (error) {
+              return <div>{error}</div>;
+            }
 
-            return <div>{data}</div>;
+            return <div>{data.toString()}</div>;
           }}
         </ResourceSubscriber>
       );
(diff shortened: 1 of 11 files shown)


src/__tests__/unit/controllers/router-store/test.tsx
```diff
@@ -474,12 +474,15 @@ describe('SPA Router store', () => {
...
       const ComponentB = () => (
         <ResourceSubscriber resource={resourceB}>
-          {({ data, loading }) => {
-            if (loading) {
-              return <div>{loading}</div>;
+          {(props) => {
+            if (props.loading) {
+              return <div>{props.loading}</div>;
+            }
+            if (props.error) {
```
theKashey, line 481: without this check accessing data will generate a complain

src/common/types.ts
```diff
@@ -66,39 +66,66 @@ export type RouteResourceLoading = boolean;
...
+export type RouteResourcePromise<T> = Promise<T> | null;
 
-export type RouteResourceUpdater = (
+export type RouteResourceUpdater<RouteResourceData = unknown> = (
   data: RouteResourceData
 ) => RouteResourceData;
 
-export type RouteResourceResponse = {
-  loading: RouteResourceLoading;
+export type RouteResourceResponseBase<RouteResourceData> = {
```
theKashey, line 79: this is a very undefined "base" shape, which might be more precisely defined later using `discriminated unions`

src/controllers/resource-utils/index.ts
```diff
@@ -1,39 +1,39 @@
...
   return async (...args: RouteResourceGettersArgs) => {
     const { default: getDataFn } = await asyncImport();
 
     return getDataFn(...args);
   };
 };
 
-export function createResource(args: CreateResourceSync): RouteResource;
-export function createResource(args: CreateResourceAsync): RouteResource;
+export function createResource<T extends unknown = RouteResourceDataPayload>(args: CreateResourceSync<T>): RouteResource<T>;
```
theKashey, line 35: note - by default `T` is equal to the previously used shape (Record) to prevent any possible breaking change.

albertogasparin:
This is great! I tried it out and seems to work very well. 
Just one small thing left: can we add proper return types to the resources examples? Currently both `homeResource` and `aboutResource` `getData` return is typed as `any`, so if we create a proper type for it we can see it "flow" into the return value of the hook

theKashey:
Sounds like you are talking about [example](https://github.com/atlassian-labs/react-resource-router/blob/2683ee17814a75262da5f53a8c73cf2eda8e69da/examples/basic-routing-with-resources/about.tsx#L8)
Sure I will update it.

> And sorry for not updating Flow part, haven't touched it for a few years and it is not soooo different

albertogasparin (approved):

src/common/types.ts
```diff
@@ -66,39 +66,66 @@ export type RouteResourceLoading = boolean;
...
+
+export type RouteResourceResponseLoaded<RouteResourceData> = {
+  loading: false;
+  error: null;
   data: RouteResourceData;
-  promise: RouteResourcePromise;
-  expiresAt: RouteResourceTimestamp;
 };
 
+export type RouteResourceResponse<
```
jackbrown, line 103: This is great!

jackbrown (approved):
@theKashey Thanks for cleaning this up!

PR #45, opened 2020-09-27
Add `basePath` support

flexdinesh:
Users can now pass in `basePath` prop and it will get appended to all paths via `<Link />`, `push` and `replace` APIs.

~**Note: Docs update yet to be done.**~ **Docs updated.**

### DIFF
--- a/src/__tests__/unit/controllers/router-store/test.tsx
+++ b/src/__tests__/unit/controllers/router-store/test.tsx
@@ -38,11 +38,13 @@ const mockHistory = {
 
 const mockRoutes = [
   {
+    basePath: '',
     path: '/pathname',
     component: () => <div>path</div>,
     name: '',
   },
   {
+    basePath: '',
     path: '/blah',
     component: () => <div>path</div>,
     name: '',
--- a/src/__tests__/unit/controllers/router-store/test.tsx
+++ b/src/__tests__/unit/controllers/router-store/test.tsx
@@ -38,13 +38,11 @@ const mockHistory = {
 
 const mockRoutes = [
   {
-    basePath: '',
     path: '/pathname',
     component: () => <div>path</div>,
     name: '',
   },
   {
-    basePath: '',
     path: '/blah',
     component: () => <div>path</div>,
     name: '',
--- a/src/common/types.ts
+++ b/src/common/types.ts
@@ -144,8 +144,6 @@ export type RouterContext = {
  * Base type for route, which doesn't contain implementation details
  */
 export type InvariantRoute = {
-  /* Base prefix that gets appended to all paths */
-  basePath?: string;
   path: string;
   exact?: boolean;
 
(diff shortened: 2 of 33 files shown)


src/common/utils/router-context/index.ts
```diff
@@ -15,10 +15,11 @@ import generatePath from '../generate-path';
 export const createRouterContext = (
```
albertogasparin, line 15: Should we change the API of this function? It already has 4 arguments and might have more later. 
I think it might be better to change it to:

```js
createRouterContext({ route, params, query, basePath })
// or
createRouterContext(route, { params, query, basePath })
```


Before it starts to get used (AFAIK no one is using it yet)

flexdinesh: I was hesitant to introduce a breaking change but`createRouterContext(route, { params, query, basePath })` seems like a good idea, especially if no one's using it yet. I'll push in the change.

flexdinesh: Should we also introduce a similar signature for `findRouterContext`? `findRouterContext(routes, { location, basePath })`

src/__tests__/unit/controllers/router-store/test.tsx
```diff
@@ -38,11 +38,13 @@ const mockHistory = {
 
 const mockRoutes = [
   {
+    basePath: '',
```
albertogasparin, line 41: This should not be here anymore, right?

src/common/types.ts
```diff
@@ -144,6 +144,8 @@ export type RouterContext = {
  * Base type for route, which doesn't contain implementation details
  */
 export type InvariantRoute = {
+  /* Base prefix that gets appended to all paths */
+  basePath?: string;
```
albertogasparin, line 148: This should not be here anymore, right?

src/utils.js.flow
```diff
@@ -20,17 +20,20 @@ declare export function generatePath(
 
 declare export function findRouterContext(
   location: Location,
-  routes: Routes
+  routes: Routes,
```
albertogasparin, line 23: Never noticed this declaration duplication and it is bad. Let's remove this utils definitions from index.flow, put the correct one here and then re-export this utils types in index

albertogasparin (changes requested):
Can we also add an integration test (for instance in `src/__tests__/unit/controllers/router/test.tsx`) to ensure that if I have `basePath = 'bla'`, route matching and history push work as expected?

src/__tests__/unit/controllers/router-store/test.tsx
flexdinesh: Yup. Removed now.

src/common/types.ts
flexdinesh: Yup. Removed now.

src/utils.js.flow
flexdinesh: Moved the duplicate declarations to util.js.flow and imported them back to index.js.flow.

src/__tests__/unit/controllers/router-store/test.tsx
flexdinesh: Created a new dir for integration tests and added two new tests. `__tests__/integration/test.tsx`

albertogasparin (dismissed):
LGTM. Should we add docs and merge?

jackbrown (approved):
Looks great

albertogasparin (approved):

PR #47, opened 2020-10-01
Use param hooks allow specifying history update type on set

billyjs:
`useQueryParam` or `usePathParam` hooks when setting the param, the update type (`push` or `update`, default: `push`) can be set to specify the how history is updated.

### DIFF
--- a/src/controllers/router-store/index.tsx
+++ b/src/controllers/router-store/index.tsx
@@ -213,7 +213,7 @@ const actions: AllRouterActions = {
     return { query, route, match };
   },
 
-  pushQueryParam: params => ({ getState }) => {
+  updateQueryParam: (params, updateType) => ({ getState }) => {
     const { query: existingQueryParams, history, location } = getState();
     const updatedQueryParams = { ...existingQueryParams, ...params };
     // remove undefined keys
@@ -228,11 +228,11 @@ const actions: AllRouterActions = {
     );
 
     if (updatedPath !== existingPath) {
-      history.push(updatedPath);
+      history[updateType ?? 'push'](updatedPath);
     }
   },
 
-  pushPathParam: params => ({ getState }) => {
+  updatePathParam: (params, updateType) => ({ getState }) => {
     const {
       history,
       location,
@@ -247,7 +247,7 @@ const actions: AllRouterActions = {
     const updatedRelativePath = getRelativeURLFromLocation(updatedLocation);
 
     if (updatedRelativePath !== existingRelativePath) {
-      history.push(updatedRelativePath);
+      history[updateType ?? 'push'](updatedRelativePath);
     }
   },
 };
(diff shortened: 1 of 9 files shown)


src/controllers/router-store/index.tsx
```diff
@@ -226,10 +226,10 @@ const actions: AllRouterActions = {
       updatedQueryParams as Query
     );
 
-    history.push(updatedPath);
+    history[updateType ?? 'push'](updatedPath);
```
jackbrown, line 229: you can just do `updateType = 'push'` on line 216

src/controllers/router-store/types.ts
```diff
@@ -61,12 +63,18 @@ type PrivateRouterActions = {
   requestRouteResources: () => RouterAction;
   listen: () => RouterAction;
   getContext: () => RouterAction;
-  pushQueryParam: (params: {
```
jackbrown, line 64: We need to be cautious when removing part of the public API. Keen for @albertogasparin to provide input here.

I feel to be safe we should keep these actions, and just proxy them through to the `update` functions. Maybe log a warning in development too.

billyjs: these are part of the private router actions, are they still part of the public API?

albertogasparin: These are private so I think it's fine

albertogasparin:
Looks good, can we update also the docs, adding this bits of API? Also can you update the flow types (index.flow)?

src/controllers/router-store/types.ts
jackbrown: ah nice, i assumed they were public.
No worries!

jackbrown:
Looks good to me, if you can update the [flow types](https://github.com/atlassian-labs/react-resource-router/blob/master/src/index.js.flow#L95) then we can get this in

flexdinesh:
Looks good @billyjs. I'm thinking a config object as the second param would make more sense but it also feels like overarching for a simple case that might never need to be extended again.

Eg. 
```
const [param, setParam] = useQueryParam('abc');

setParam('hello', { method: 'PUSH' });
```

Config object can be 
```
type config = { 
 method: 'PUSH' | 'REPLACE'
}
```

But if we think deeply about it, this definitely seems like over-complicating a simple case. Just leaving my thoughts here for feedback from other maintainers. @albertogasparin @jackbrown

jackbrown:
@flexdinesh I don't think this is necessary for now as we are only planning to add the single argument. We can always extend the API in future if we feel there are more arguments required

jackbrown (approved):
Looks good to me @billyjs !

flexdinesh (approved):

albertogasparin:
Can we resolve the conflicts and get this in today?

jackbrown (approved):

albertogasparin (approved):

PR #48, opened 2020-10-08
prevent rerenders if hook setters are called with the same value

jackbrown:
I noticed in the examples that if you click on the "remove param" buttons on the hooks multiple times, the page continues to rerender even though the path hasn't updated.

This PR changes the behavior so we only call `history.push` if the path has changed.

### DIFF
--- a/src/controllers/router-store/index.tsx
+++ b/src/controllers/router-store/index.tsx
@@ -227,8 +227,6 @@ const actions: AllRouterActions = {
       updatedQueryParams as Query
     );
 
-    console.log('updatedPath', updatedPath);
-
     if (updatedPath !== existingPath) {
       history.push(updatedPath);
     }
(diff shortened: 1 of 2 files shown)


src/controllers/router-store/index.tsx
```diff
@@ -221,12 +221,17 @@ const actions: AllRouterActions = {
...
         updatedQueryParams[key] === undefined && delete updatedQueryParams[key]
     );
+    const existingPath = updateQueryParams(location, existingQueryParams);
     const updatedPath = updateQueryParams(
       location,
       updatedQueryParams as Query
     );
 
-    history.push(updatedPath);
+    console.log('updatedPath', updatedPath);
```
albertogasparin, line 230: 👻

albertogasparin (changes requested):

src/controllers/router-store/index.tsx
jackbrown: 🤦 removed

albertogasparin (approved):

flexdinesh (approved):

PR #59, opened 2020-12-11
Add createRouterSelector API

### LINKED ISSUE #57: Allow creating custom hooks to reduce `useRouter` re-renders
Using `useRouter` everywhere is suboptimal as triggers a re-renders even when we might just care about `route.name`.
So we should allow consumers to create their own optimised selector hooks:

```js
const useRouteName = createRouterSelector((o) => o.route.name);
```

albertogasparin:
Expose a `createRouterSelector` that returns a hook to access a selected part of the router store.

```js
const useRouteName = createRouterSelector(s => s.route.name);
```

Closes #57

### DIFF
--- a/src/controllers/router-store/index.tsx
+++ b/src/controllers/router-store/index.tsx
@@ -26,6 +26,7 @@ import {
   ContainerProps,
   UniversalRouterContainerProps,
   EntireRouterState,
+  RouterState,
 } from './types';
 
 import { Query } from '../../common/types';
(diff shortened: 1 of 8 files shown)


src/controllers/router-store/index.tsx
```diff
@@ -360,6 +361,20 @@ export const useRouterStoreStatic = createHook<
   selector: null,
 });
 
+/**
+ * Utility to create custom hooks subscribing to   without re-rendering on route change
```
MonicaOlejniczak, line 365: Can you fix the spacing on `to   without`

jackbrown (approved):
This looks great!

MonicaOlejniczak (approved):
LGTM

PR #64, opened 2020-12-31
Remove deprecated type from withRouter

JakeLane:
Fixing the flowtype for withRouter to remove implicit any, which fixes props not being checked on HOC.

https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-router-dom_v5.x.x/flow_v0.98.x-v0.103.x/react-router-dom_v5.x.x.js#L155


jackbrown (approved):
LGTM

PR #65, opened 2021-01-02
Add resource prefetch support

### LINKED ISSUE #61: Add preload/prefetch API on Link hover
We should support resources to be fetched on hover. Its also needs to 
- handle maxAge properly (avoid double fetch)
- provide an AbortController so requests can be dropped
- expose API so consumers can to run custom code (to preload RLL bundles for instance)
- this behaviour should be optional, with a customisable delay

albertogasparin:
Closes #61 

Add new prop to `Link`: `prefetch` that can be used to prefetching resources ahead of time.
Supported values are: `false`, `mount` and `hover`. 
Prefetch is triggered after a delay of 300ms. Also, we force resource `maxAge` to be 10s regardless of what the real value is, so we allow even resources with `maxAge: 0` to be prefetched.

Resource router context now also has a `isPrefetch` boolean so resources can decide to handle prefetch differently

We also add a new prop to router `onPrefetch` so we can trigger other side effects when prefetch triggers (like preloading js bundles)


src/ui/link/index.tsx
```diff
@@ -49,7 +68,28 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
...
+      if (typeof to !== 'string' && !route) return;
+
+      const context =
+        typeof to !== 'string' && route
+          ? createRouterContext(route, { params, query })
+          : null;
+      routerActions.prefetchNextRouteResources(linkDestination, context);
+      // omit params & query as already in linkDestination
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [route, linkDestination, routerActions]);
```
MonicaOlejniczak, line 82: Shouldn't `to` be in here?

src/ui/link/index.tsx
```diff
@@ -49,7 +68,28 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
           ''
         : to;
 
-    const handleLinkPress = (e: any) =>
+    const triggerPrefetch = useCallback(() => {
+      // ignore if async route not ready yet
+      if (typeof to !== 'string' && !route) return;
+
+      const context =
+        typeof to !== 'string' && route
```
MonicaOlejniczak, line 76: Shouldn't this condition always be true with the early return above?

src/ui/link/index.tsx
```diff
@@ -59,6 +99,20 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
         to: route && [route, { params, query }],
       });
 
+    const handleMouseEnter = (e: MouseEvent) => {
+      if (prefetch === 'hover') {
+        prefetchRef.current = setTimeout(triggerPrefetch, PREFETCH_DELAY);
```
MonicaOlejniczak, line 104: /nit The ref can also be cleared within the `setTimeout` alongside the existing `onMouseLeave` handler

src/controllers/router-store/index.tsx
```diff
@@ -135,6 +136,35 @@ const actions: AllRouterActions = {
...
+      getContext: getResourceStoreContext,
+    } = getResourceStore().actions;
+
+    if (!nextContext && !isExternalAbsolutePath(path)) {
+      const location = parsePath(getRelativePath(path, basePath) as any);
+      nextContext = findRouterContext(routes, { location, basePath });
+    }
+
+    if (nextContext == null) return;
+    const nextLocationContext = nextContext;
```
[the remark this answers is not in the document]
albertogasparin: Apparently TS complains, as the `null` refinement does not properly work 😞

src/ui/link/index.tsx
albertogasparin: Yes, but TS complains saying route could be null 😞

src/ui/link/index.tsx
albertogasparin: `linkDestination` covers that too, as it is a function of `to`. Will add to the comment

jackbrown:
How are we planning to implement this for routes that share resources? If I use a resource on both route A and B, then page A will rerender when B is prefetched.

Is this something you think the router should solve, or should direction be to only share resources if the `maxAge` is set to `Infinity`?

albertogasparin:
> How are we planning to implement this for routes that share resources? If I use a resource on both route A and B, then page A will rerender when B is prefetched.

`getResourcesForNextLocation` already takes care of this eventuality. We filter and do not refetch resources that are shared regardless of expiration status (the same behaviour of normal route change)

MonicaOlejniczak (approved):

PR #66, opened 2021-01-18
Add initialRoute to Router flow type

liamqma:
Fixing Router flow type by adding InitialRoute


albertogasparin:
Already fixed in #65. Thanks anyway

PR #71, opened 2021-02-15
node 15

taoqf:


albertogasparin:
Thanks for the submission! We now officially support node 16, so 15 is good too 😊

PR #82, opened 2021-04-27
Add max wait time as an option to static router

CharlesCarey:


albertogasparin:
I think we are missing Flow types update, right?

src/controllers/resource-store/utils/generate-time-guard/index.ts
```diff
@@ -0,0 +1,21 @@
+import { GenerateTimeGuardReturn } from './types';
+
+export const generateTimeGuard = (
```
albertogasparin, line 3: Can we add two specific unit tests for this funciton, just to ensure it works as expected (returned loading state and resolved state)

src/controllers/static-router/index.tsx
```diff
@@ -46,7 +46,7 @@ StaticRouter.requestResources = async (props: RequestResourcesParams) => {
 
   bootstrapStore({ ...bootstrapProps, ...overrides });
 
-  await requestRouteResources();
+  await requestRouteResources({ maxWaitTime });
```
albertogasparin, line 49: I think this is called also from universal router

jackbrown:
This is looking great.

Can you help me understand the use case? Is this intended to help the app timeout in SSR sooner than it would if it waited for the actual SSR timeout? Is there any plan to use different timeouts on different resources?

src/__tests__/unit/controllers/static-router/test.tsx
```diff
@@ -116,6 +116,40 @@ describe('<StaticRouter />', () => {
       });
     });
 
+    it('should respect max wait time when fetching resources', async () => {
```
albertogasparin, line 119: 'should respect timeout when fetching resources'

src/__tests__/unit/controllers/resource-store/test.tsx
```diff
@@ -138,6 +139,28 @@ describe('resource store', () => {
         });
       });
 
+      it('should respect max wait time option', async () => {
```
albertogasparin, line 142: 'should respect timeout option'

src/__tests__/unit/controllers/static-router/test.tsx
```diff
@@ -116,6 +116,40 @@ describe('<StaticRouter />', () => {
...
+            data: 'hello world',
+            error: null,
+            loading: false,
+            promise: null,
+            expiresAt,
+          },
+        },
+      });
+    });
+
```
albertogasparin, line 152: Can we add another integration test: `should re-trigger requests for timed out resources when mounted` ?

docs/router/ssr.md
```diff
@@ -73,4 +73,6 @@ const renderToStringWithData = async ({ location }) => {
 };
 ```
 
-Notice that we do not need to provide any `resourceData` object to the `ServerApp`,the `StaticRouter` handles this for us internally.
+Notice that we do not need to provide any `resourceData` object to the `ServerApp`, the `StaticRouter` handles this for us internally.
+
+To prevent slow APIs from causing long renders on the server you can optionally pass in `timeout` as an option to `StaticRouter.requestResources`. If a route resource does not return within the specified time then its data and promise will be set to null.
```
albertogasparin, line 78: We should add somewhere (in the `useResource` docs) a table about combinations of states:

| state | data | loading | error | promise |
|------|------|--------|------|------------|
| idle | `null` | `false` | `null` | `null` |
| loading | `null` or `{}` | `true` | `null` or `Error` | `Promise` |
| loading after ssr timeout | `null` | `true` | `TimeoutError` | `null` |
| fetch successful  | `{}` | `false` | `null` | `Promise` |
| fetch error   | `null` or `{}` | `false` | `Error` | `Promise` |

Also, we should add an explicit note in `useResource` docs saying that in SSR, `loading` can be `true` even when there is an error, but promise will be `null` because there is no `Suspense` support. So devs should give priority to `loading` to decide between loading or error state for their components, and only throw promise/error on the client

albertogasparin (changes requested):

albertogasparin:
> Can you help me understand the use case? Is this intended to help the app timeout in SSR sooner than it would if it waited for the actual SSR timeout? Is there any plan to use different timeouts on different resources?

Yes, actual SSR timeout currently does not render anything and loses data already fetched. We want to instead render a skeleton for components that have slow APIs but still render others that are ready. So SSR timeout should kinda never be hit (unless we have an issue with render or something).

> Is there any plan to use different timeouts on different resources? 
Maybe, so far we did not have any specific request to support for it thought

CharlesCarey:
> This is looking great.
> 
> Can you help me understand the use case? Is this intended to help the app timeout in SSR sooner than it would if it waited for the actual SSR timeout? Is there any plan to use different timeouts on different resources?

@jackbrown the idea behind this change is to make sure we can always SSR some sort of content to the client. We have a hard timeout around the total SSR time already, if we surpass this then we throw away all of the SSRed result so the user sees a blank screen. With this change we can impose a limit on how long resources can take to fetch, leaving us enough time to complete the render. We should always be able to send something meaningful to the client, even if it is just a loading state. 

The change will accomplish two things. Firstly, we can improve the user experience in the case of slow APIs (loading state vs blank screen) and we will be able to improve overall performance as we will only have to refetch failed resources on the client rather than everything.

jackbrown:
Really great stuff. We should share this feature around to all consumers of RRR and SSR once it's available!

albertogasparin (approved):

PR #83, opened 2021-04-27
Add generics for typing resource data in Flow types

nickrobson:
Right now, our Flow types don't make type guarantees for consumers of resources via `useResource` and `ResourceSubscriber`, because resource definition (and hence consumption) has no way of providing the response shape to Flow.

This PR addresses this issue, by adding generic parameters to `createResource` which are then passed through for inference by `useResource` and `ResourceSubscriber`. Please see comments on relevant places for context/discussion points. :)

----

Example diff of a consumer upgrading to consume this:
```diff
 // @flow
 import { createResource, useResource } from 'react-resource-router';
 
 type ResponseType = {
   ok: boolean,
   value: number,
 }

-const helloWorldResource = createResource({
+const helloWorldResource = createResource<ResponseType>({
   type: 'hello-world',
   getKey: () => 'global (no pun intended)',
   getData: () => Promise.resolve({ ok: true, value: 9001 }),
 }

 const Component = () => {
   const { data } = useResource(helloWorldResource);
   const value = data?.value;
-  // Flow thinks `data` is `any` so `value` is now `any` too!
-  const stringValue: string = value; // wait what?!
+  const stringValue: string = value; // Flow now errors here <3
 }
```

### DIFF
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -42,14 +42,14 @@ declare export function Router(props: RouterProps): Node;
 declare export function MemoryRouter(props: MemoryRouterProps): Node;
 declare export function StaticRouter(props: StaticRouterProps): Node;
 
-declare export function ResourceSubscriber(props: {
+declare export function ResourceSubscriber<T>(props: {
   children: (
-    resource: RouteResourceResponse & {
-      update: (getNewData: RouteResourceUpdater<RouteResourceData>) => void,
+    resource: RouteResourceResponse<T> & {
+      update: (getNewData: RouteResourceUpdater<RouteResourceData<T>>) => void,
       refresh: () => void,
     }
   ) => Node,
-  resource: RouteResource,
+  resource: RouteResource<T>,
 }): Node;
 declare export function RouterSubscriber(props: RouterSubscriberProps): Node;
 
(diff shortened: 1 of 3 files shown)


src/index.js.flow
```diff
@@ -120,24 +120,24 @@ declare export function createRouterContext(
...
+        RouteResource<T>,
+        {| getData: $PropertyType<RouteResource<T>, 'getData'> |}
       >,
       maxAge?: number,
-      getDataLoader: GetDataLoader,
+      getDataLoader: GetDataLoader<T>,
     |};
 
-declare export function createResource(args: CreateResourceArg): RouteResource;
+declare export function createResource<T>(args: CreateResourceArg<T>): RouteResource<T>;
```
nickrobson, line 143: Note that this makes `createResource` _require_ a type argument. It won't default to `any` and Flow doesn't seem to be able to infer the type based on `getData`/`getDataLoader`. However, the conversion is fairly straight forward; it just takes the type of the result that `getData`'s returned Promise returns.

albertogasparin (approved):

src/types.js.flow
```diff
@@ -124,7 +126,7 @@ export type Route = {
   /**
    * The resources for the route
    */
-  resources?: RouteResource[],
+  resources?: RouteResource<any>[],
```
MonicaOlejniczak, line 129: Would mixed be better here? Though I know flow handles it weirdly, so it might be painful to actually use

nickrobson: Good callout! I did try to use `mixed` and `empty` here but unfortunately neither worked without causing further pain when integrating these changes into Jira's frontend. 😓  I suspect it'd need further work to make it work nicely.

jackbrown (approved):

PR #84, opened 2021-04-30
correction for docs to go correct page instead of 404

mahmood-sajjadi:
Some of the links in docs are going to 404 due to missing folder/extension name in url or typos.

### DIFF
--- a/docs/_sidebar.md
+++ b/docs/_sidebar.md
@@ -1,22 +1,22 @@
 - **Resources**
 
-  - [Concept](/resources/concept.md)
-  - [Creation](/resources/creation.md)
-    - [Code Splitting](/resources/creation/code-splitting.md)
-    - [Error Handling](/resources/creation/error-handling.md)
-  - [Adding](/resources/adding.md)
-  - [Usage](/resources/usage.md)
-  - [Interaction](/resources/interaction.md)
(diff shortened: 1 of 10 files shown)


mahmood-sajjadi:
Ok I revert some of my changes as I found this link `https://atlassian-labs.github.io/react-resource-router/#/` from Atlassian internal resources (I am an Atlassian developer)
But this link is not available in npm or repo itself.
I am suggesting to change the link in npm `https://www.npmjs.com/package/react-resource-router` for Homepage to docs links otherwise none of the docs link in github will work.
If you open the docs in github itself, all of the links will go to 404 as /docs/ and extension (.md) is missing from the urls

albertogasparin:
Thanks for the fix! Yeah, supporting both .md navigation and website navigation would be nice. By looking at how we did it in https://github.com/atlassian/react-sweet-state/tree/master/docs it might be a matter of converting all `/bla/foo` to be relative (`bla/foo`). 
Would you like to give it a go?

mahmood-sajjadi:
sure that would be an honor. I will update this PR to be the relative link

mahmood-sajjadi:
Done, all of them links in PR changed to be relative links.

albertogasparin (approved):

Issue #104, opened 2021-11-04
Same requests are sent multiple time if previous one not resolved


chloeleichen:
See an example with jira-dashboard
https://user-images.githubusercontent.com/7508939/140255451-a957acab-0693-40ec-9de5-42cccd8f3a3e.mov

jackbrown:
Thanks @chloeleichen for the report. I can replicate the issue in this sandbox https://codesandbox.io/s/rrr-replicating-issue-104-lqm0m?file=/src/about/index.tsx

albertogasparin:
Fixed by #115 and release in `v0.19.0`

PR #105, opened 2021-11-25
ensure consistent resource promise

bholloway:
Best reviewed commit at a time.

Primarily I want to ensure that the resource `promise` will always resolve `data` or reject `error`. The main culprit being `update()` where the `promise` becomes immediately mismatched with the `data`.

Timeouts should also be rejecting errors IMO but are not included here. The logic to do that is not really more complicated  but there is a more general problem that `expiresAt` and `promise` need to be set during hydration rather than being mutated on first access. I attempted this refactor but ultimately got stuck on the tests.

Additionally in the remote resource fetch
* I made the the different async results more declarative for better clarity
* covered the edge case where an explicit update aborts in-flight fetch.

I did a little work to make the tests more consistent but there is still a lot that needs doing there.

### DIFF
--- a/src/controllers/resource-store/index.tsx
+++ b/src/controllers/resource-store/index.tsx
@@ -49,11 +49,7 @@ export const actions: Actions = {
     getState,
     dispatch,
   }) => {
-    const { data: resourceStoreData } = getState();
-    const slice = getSliceForResource(
-      { data: resourceStoreData },
-      { type, key }
-    );
+    const slice = getSliceForResource(getState(), { type, key });
     const data = getNewSliceData(slice.data);
 
     dispatch(
@@ -79,12 +75,9 @@ export const actions: Actions = {
   }) => {
     const { type, getKey, maxAge } = resource;
     const { getResourceFromRemote } = actions;
-    const { data: resourceStoreData, context } = getState();
+    const { context, ...resourceStoreState } = getState();
     const key = getKey(routerStoreContext, context);
-    let cached = getSliceForResource(
-      { data: resourceStoreData },
-      { type, key }
-    );
+    let cached = getSliceForResource(resourceStoreState, { type, key });
 
     if (shouldUseCache(cached)) {
       if (isFromSsr(cached)) {
@@ -111,12 +104,12 @@ export const actions: Actions = {
   }): Promise<RouteResourceResponse<unknown>> => {
     const { type, getKey, getData, maxAge } = resource;
     const { prefetch, timeout } = options;
-    const { data: resourceStoreData, context } = getState();
+    const { context, ...resourceStoreState } = getState();
     const key = getKey(routerStoreContext, context);
-    const slice = getSliceForResource(
-      { data: resourceStoreData },
-      { type, key }
-    );
+    const slice = getSliceForResource(resourceStoreState, {
+      type,
+      key,
+    });
 
     if (slice.loading) {
       return slice;
--- a/src/controllers/resource-store/index.tsx
+++ b/src/controllers/resource-store/index.tsx
@@ -49,17 +49,24 @@ export const actions: Actions = {
     getState,
     dispatch,
   }) => {
-    const { data } = getState();
-
-    const slice = getSliceForResource({ data }, { type, key });
+    const { data: resourceStoreData } = getState();
+    const slice = getSliceForResource(
+      { data: resourceStoreData },
+      { type, key }
+    );
+    const data = getNewSliceData(slice.data);
 
     dispatch(
-      setResourceState(type, key, {
-        ...slice,
-        data: getNewSliceData(slice.data),
-        expiresAt: getExpiresAt(maxAge),
-        accessedAt: getAccessedAt(),
-      })
+      data == null
+        ? deleteResourceKey(key, type)
+        : setResourceState(type, key, {
+            ...slice,
+            data,
+            error: null,
+            promise: Promise.resolve(data),
+            expiresAt: getExpiresAt(maxAge),
+            accessedAt: getAccessedAt(),
+          })
     );
   },
   /**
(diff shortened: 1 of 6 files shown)


jackbrown (commented):
@bholloway does this fix #104?

bholloway:
@jackbrown not specifically.

This is to fix a conceptual correctness problem - IMO the `promise` should resolve `data` or `error` and never stale `data`. Specifically `update(() => null)` to delete an item should not keep a reference to data in the promise.

In react the `promise` resolved value should not really be used, there are better ways for observing resource `data`. However in playing with some "dependent resources" implementation outside react I find the promise is needed as an indicator of future value.

Since I started the branch I think this correctness with the `promise` was partly addressed during the LRU changes so ultimately there was not a lot to do. Although I was tempted to make other fixes to tests etc which would be good to land. All that is to say we should discuss offline when/how these changes should land.

src/controllers/resource-store/index.tsx
```diff
@@ -49,17 +49,20 @@ export const actions: Actions = {
...
+    const data = getNewSliceData(slice.data);
 
     dispatch(
-      setResourceState(type, key, {
-        ...slice,
-        data: getNewSliceData(slice.data),
-        expiresAt: getExpiresAt(maxAge),
-        accessedAt: getAccessedAt(),
-      })
+      data == null
```
liamqma, line 56: We are currently working on treating `null` as a valid data. Sorry for making things complicated but I am unsure that resource is removable if data is null.

bholloway: In that case it's easy to not delete the record here and just treat `null` as any other update. Either way the `promise` should resolve the `null` for consistency, which is the purpose of my PR. That said I think there needs to be some seperate semantics for a `delete()` operation.

However the bigger problem I see with `null` being valid data is that it is indistinguishable from the initial state. That is a big problem IMO for determining what state the resource is in.

liamqma: > However the bigger problem I see with null being valid data is that it is indistinguishable from the initial state. That is a big problem IMO for determining what state the resource is in.

Very true. I am examining different options. One of the solutions i think of is to set initial value to be 
```
export const BASE_DEFAULT_STATE_SLICE = {
  data: "DEFAULT_STATE_DATE",
};
```
in order to be distinguishable
What do you think?

bholloway: hmm not a fan, it needs to be a primitive type.

Really this is the whole point of `undefined` IMO.

bholloway: I think that combined with a presence of `promise` might be enough. I would need to look at the table in the `.md` file. But certainly we need to be very precise about that table going forward.

IIRC Apollo had some `linkState`(?) which makes the state explicit. I was always a fan of that since it avoids this type of problem. However I think that such a change is a bit extreme given the current API doesn't have that.

src/controllers/resource-store/index.tsx
```diff
@@ -104,75 +104,99 @@ export const actions: Actions = {
...
+          };
         } else {
           timeoutGuard.timerId && clearTimeout(timeoutGuard.timerId);
-          response.data = maybeData;
-          response.loading = false;
+          resolvedSlice = {
+            data: maybeData,
+            error: null, // previous error is removed
+            loading: false,
+            promise,
```
albertogasparin, line 157: `pendingSlice` already has `promise` being set. Is there any value setting it to `resolvedSlice` ?

src/controllers/resource-store/index.tsx
```diff
@@ -104,75 +104,99 @@ export const actions: Actions = {
...
+        error,
+        loading: false,
+      };
     }
 
-    response.expiresAt = getExpiresAt(
-      options.prefetch && maxAge < PREFETCH_MAX_AGE ? PREFETCH_MAX_AGE : maxAge
-    );
+    // do not overwrite a slice that was updated during async
+    const laterSlice = getSliceForResource(getState(), { type, key });
```
albertogasparin, line 178: Are we sure about this? If I'm implementing some sort of optimistic updates, this will discard server data for client-side data that I might have manually set, and which might be incorrect (especially if server returns an error).

bholloway: IMO any optimistic update needs to be done before refresh otherwise it implicitly relies on the async nature of refresh to avoid race condition.

Fundamentally if the `promise` has changed during the async we are in a difference state.

The options as I see it are
1. explicit updates do no effect the `promise`, and the `promise` mismatches `data` (current state before PR)
2. explicit updates set the `promise` consistent with `data`, but preceding async refresh will set it back when it completes
3. explicit updates set the `promise` consistent with `data`, cancels in-flight async refresh

Fixing (1) is the point of this PR so I'd say its a choice between (2) and (3).
Choosing (2) puts us in a position where the `promise` no longer reflects the future value. So is also inconsistent in its own way.
I would specifically chose (3) to force developers to consider the order. It is the only solution that makes `promise` consistent with current `data` and future async `data`.

src/controllers/resource-store/index.tsx
bholloway: IMO yes. In this block there is a branch where a timeout will produce `promise: null`. Without declaring the value of `promise` in each branch it is harder to compare the final state at a glance.

albertogasparin: Then, should we go all explicit and no longer spread `pendingSlice` in `finalSlice` ? As I'd say either we leverage spreading for all non changing data, or we explicit set the state for every step. Seems like the latter is preferable for clarity, so I'd enforce it

src/controllers/resource-store/index.tsx
albertogasparin: > IMO any optimistic update needs to be done before refresh otherwise it implicitly relies on the async nature of refresh to avoid race condition.

Agree. That is why I would make the async flow win anyway.

> Fundamentally if the promise has changed during the async we are in a difference state.

I'm confused on when this might happen. If we exclude a manual update, then there should be no way for something else to change that promise that is not this async flow itself. 
If this is a way to mitigate #104, then I think we need to fix that bug and not account for that eventuality at all.

bholloway:
I'll drop this because of the conflict with the browser-only change.

Good discussion about race conditions with refresh. I will raise a separate PR with a more refined approach.

PR #108, opened 2021-12-21
Allow Redirect component to accept Route object as `to`

jpqne:
This would allow the `Redirect` component to accept a route object as a prop, similar to `Link`.

### DIFF
--- a/src/controllers/redirect/index.tsx
+++ b/src/controllers/redirect/index.tsx
@@ -6,6 +6,7 @@ import { RouterActionsType, RouterState } from '../router-store/types';
 import { RouterSubscriber } from '../subscribers/route';
 
 import { RedirectProps } from './types';
+import { generateLocationFromPath } from '../../common/utils';
 
 type RedirectorProps = RedirectProps & {
   actions: RouterActionsType;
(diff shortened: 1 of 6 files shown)


src/controllers/redirect/index.tsx
```diff
@@ -18,8 +19,18 @@ class Redirector extends Component<RedirectorProps> {
   };
 
   componentDidMount() {
-    const { to, location, push, actions } = this.props;
-    const newPath = typeof to === 'object' ? createPath(to) : to;
+    const { to, location, push, params, query, actions } = this.props;
+    const routeAttributes = {
+      params,
+      query,
+      basePath: actions.getBasePath() as any,
```
JakeLane, line 26: Is there a reason the type is casted to any here?

src/controllers/redirect/types.ts
```diff
@@ -1,6 +1,8 @@
-import { Location } from '../../common/types';
+import { Location, MatchParams, Query, Route } from '../../common/types';
 
 export type RedirectProps = {
-  to: Location | string;
+  to: Location | Route | string;
```
JakeLane, line 4: I think the flow type should be updated too

JakeLane (changes requested):

albertogasparin (dismissed):

JakeLane (dismissed):

liamqma (approved):

liamqma (approved):

JakeLane (approved):

Issue #111, opened 2022-01-27
react-resource-router does not render matched component on redirect or hash change with hashrouter


spendo-atl:
If you use createHashHistory with react-resource-router like so:
```
<Router routes={routes} history={createHashHistory()}>
        <RouteComponent />
</Router>
```

When you change the url hash through <Redirect/> or manually it does not render anything (or the matching route component). 

When reloading the page with the correct hash route it then matches and renders the correct component.

We've created a minimal reproducible example below:
https://codesandbox.io/s/react-resource-router-hash-router-bug-lmzfe

albertogasparin:
Seems like the issue is about the actual support of history v5, that currently does not work at all

PR #113, opened 2022-01-27
Ensure compatibility with history@5 #111

marionebl:


src/__tests__/unit/controllers/router-store/test.tsx
```diff
@@ -243,6 +244,98 @@ describe('SPA Router store', () => {
     });
   });
 
+  describe.only('listening for real history changes on history@5', () => {
```
albertogasparin, line 247: Please remove `.only`

albertogasparin:
@liamqma This looks good to me, but do you think we need more changes?

src/common/utils/create-location/parse-path.ts
```diff
@@ -0,0 +1,13 @@
+export function parsePath(path: string) {
+  const url = new URL(path, 'ws://a.a');
```
albertogasparin, line 2: Out of curiosity, why `ws://a.a` ?

albertogasparin (approved):

src/common/utils/create-location/parse-path.ts
marionebl: Its the shortest url I could come up with 😅

package.json
```diff
@@ -131,6 +132,7 @@
     "webpack-dev-server": "^3.11.0"
   },
   "peerDependencies": {
+    "history": "^4 || ^5",
```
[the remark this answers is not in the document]
marionebl: Yes, leaning on https://semver.npmjs.com/ I think they are equivalent:

Screen Shot 2022-02-15 at 10 00 31 am
Screen Shot 2022-02-15 at 10 00 19 am

Issue #106, opened 2021-12-02
Links with preload trigger fetch request even for resources on current route


albertogasparin:
It might be an issue with expired resources.

Load a next-gen board, then open your work - boards and hover over that same board link. You'll see the board refreshing

liamqma:
yup, i can reproduce this issue.

PR #125, opened 2022-02-17
update routeHasChanged() to check route name and group instead of object reference

liamqma:
Update routeHasChanged() to check if both route name and group match rather than to compare them by ref.

This is to fix https://github.com/atlassian-labs/react-resource-router/issues/106,

We might have something like:
```
<Link to={import(/* webpackChunkName: "issue-route" */'./issue-route')} prefetch="hover"></Link>
```
I guess comparing object reference isn't bulletproof and the combination of route name and group should be unique.

### DIFF
--- a/src/controllers/resource-store/utils/route-checks/index.ts
+++ b/src/controllers/resource-store/utils/route-checks/index.ts
@@ -4,4 +4,4 @@ export const routeHasResources = (route: Route | null): boolean =>
   !!(route && route.resources && route.resources.length > 0);
 
 export const routeHasChanged = (prev: Route, next: Route): boolean =>
-  prev.name !== next.name || prev.group !== next.group;
+  prev.name !== next.name;
--- a/src/controllers/resource-store/utils/route-checks/index.ts
+++ b/src/controllers/resource-store/utils/route-checks/index.ts
@@ -4,4 +4,4 @@ export const routeHasResources = (route: Route | null): boolean =>
   !!(route && route.resources && route.resources.length > 0);
 
 export const routeHasChanged = (prev: Route, next: Route): boolean =>
-  prev.name !== next.name;
+  prev.name !== next.name || prev.path !== next.path;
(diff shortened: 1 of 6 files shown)


src/controllers/resource-store/utils/route-checks/index.ts
```diff
@@ -3,7 +3,5 @@ import { Route } from '../../../../common/types';
 export const routeHasResources = (route: Route | null): boolean =>
   !!(route && route.resources && route.resources.length > 0);
 
-export const routeHasChanged = (
-  prev: Route | null,
-  next: Route | null
-): boolean => prev !== next;
+export const routeHasChanged = (prev: Route, next: Route): boolean =>
+  prev.name !== next.name || prev.group !== next.group;
```
JakeLane, line 7: `name` should always be unique so checking `group` is an unnecessary step

JakeLane (approved):

JakeLane (dismissed):

src/controllers/resource-store/utils/route-checks/index.ts
albertogasparin: While I agree, given we do not enforce unique names, I wonder if we should have `name` + `path` check (just to be on the safer side). Also `group` is something we use internally but never documented. Just noticed how `Route` type also has `canTransitionIn/Out`, something we never ported

albertogasparin (approved):

src/controllers/resource-store/utils/route-checks/index.ts
liamqma: Sounds good. I've added `name` + `path` check. Also, removed canTransitionIn/Out and group from types.
I noticed that type `JiraRoute` doesn't have `group` and `canTransitionIn`. I guess when we bump RRR in Jira, we will remember to update `JiraRoute`. Flow'll complain anyway 😀.

JakeLane (approved):

PR #145, opened 2022-04-02
Consolidate Router component

albertogasparin:
We currently have 4 router components, for no particular reason. They indeed support different features as they have been poorly updated across evolution. This is an initial step toward simplification.

Notable changes:
- remove `UniversalRouter` and `StaticRouter`. `Router` now does what they did, passing the same props
- `MemoryRouter` is now forced local, as persisted state cause more trouble than anything in tests
- `isStatic` prop is gone, we detect node to skip `isBrowserOnly` resources


docs/router/ssr.md
```diff
@@ -22,22 +22,25 @@ import { RouteComponent } from 'react-resource-router';
...
   <>
+    <Navigation />
     <RouteComponent />
+    <Footer />
   </>
 );
 ```
 
-The reason for this is that currently, you will need to use the [`Router`](#router-component) component on the client and the [`StaticRouter`](#staticrouter-component) component on the server. Following the above composition pattern will allow you to use the correct router in your server side entry and client side entry respectively. This could look something like the following examples:
+When however you need to SSR your app, we need pass different props to Router, as `createBrowserHistory` does not really work on server, so we recommend to use `location` instead (or pass your own `MemoryHistory` if needed)
```
liamqma, line 32:
```suggestion
When you need to SSR your app, we need to pass different props to Router, as `createBrowserHistory` does not really work on server, so we recommend to use `location` instead (or pass your own `MemoryHistory` if needed)
```

liamqma (dismissed):

liamqma (approved):

PR #151, opened 2022-05-04
No longer recreating histories within MemoryRouter

PatrickShaw:
Similar to #150 I don't think we want to create a new memory history on every render of `MemoryRouter`. Memory history is supposed to maintain state outside the lifetime of a single render.

I've fixed this issue by putting `useRef` around the `createMemoryHistory` to retain the same instance.


src/controllers/memory-router/index.tsx
```diff
@@ -1,57 +1,73 @@
...
+    }
+
+    return current;
   };
+};
 
-  if (resourceData) {
-    routerProps = { ...routerProps, resourceData };
-  }
+const useMemoryHistory = (location: string | undefined) => {
```
PatrickShaw, line 23: Stored all the ugliness of trying to retain the same instance across renders in this hook.

PR #152, opened 2022-05-06
Improvements to consistency

bholloway:
Support for synchronous return from `getData()`.

Fix some inconsistencies (potential bugs as marked).

```
actions
  getResourceFromRemote
    where getData() synchronously returns new data
      before:
        ✓ should enter loading state with new data as promise [BUG] (prefetch=false)
        ✓ should enter loading state with new data as promise [BUG] (prefetch=true)
      after:
        ✓ should enter loading state with promise resolving new data (prefetch=false)
        ✓ should enter loading state with promise resolving new data (prefetch=true)
    where getData() times out
      before:
        ✓ should enter loading state with promise unrepresentive of rejecting TimeoutError [BUG] and clear existing data on loaded [BUG] (prefetch=false, w/ timeout)
        ✓ should enter loading state with promise unrepresentive of rejecting TimeoutError [BUG] and clear existing data on loaded [BUG] (prefetch=true, w/ timeout)
      after:
        ✓ should enter loading state with promise rejecting TimeoutError and preserve existing data on loaded (prefetch=false, w/ timeout)
        ✓ should enter loading state with promise rejecting TimeoutError and preserve existing data on loaded (prefetch=true, w/ timeout)
consistency of promise and data under race conditions
  when update occurs outside loading
    before:
      ✓ should retain stale error [BUG] and stale promise [BUG] on update
      ✓ should set promise when loading but keep data and error (assume not expired)
      ✓ should clear error and set new data and on loaded
    after:
      ✓ should clear error and set promise on update
      ✓ should set promise when loading but keep data (assume not expired)
      ✓ should set new data and on loaded
  when update occurs during loading
    before:
      ✓ should set promise when loading but retain data and error (assume not expired)
      ✓ should retain stale error [BUG] and async promise on update
      ✓ should clear error and set new data on loaded
      ✓ should reinstate stale initial data [BUG] and set error on failed
    after:
      ✓ should set promise on loading but retain data and error (assume not expired)
      ✓ should clear error on update but retain promise
      ✓ should set new data on loaded
      ✓ should retain updated data and set error on failed
  when clear occurs during loading
    before:
      ✓ should set promise on loading but retain data and error (assume not expired)
      ✓ should delete resource state on clear
      ✓ should not reinstate resource slice on loaded
      ✓ should not reinstate resource slice on failed, should reinstate stale data [BUG]
    after:
      ✓ should set promise on loading but retain data and error (assume not expired)
      ✓ should delete resource state on clear
      ✓ should not reinstate resource slice on loaded
      ✓ should not reinstate resource slice on failed

```


albertogasparin (dismissed):

src/controllers/resource-store/index.tsx
```diff
@@ -89,101 +135,205 @@ export const actions: Actions = {
...
+      return Promise.resolve(cached);
     }
 
     return dispatch(
-      getResourceFromRemote(resource, routerStoreContext, options)
+      privateActions.getResourceFromRemote(
+        resource,
+        routerStoreContext,
+        options
+      )
```
bholloway, line 146: On reflection I think we should revert changes to this function because of how it effects getting all resources on the StaticRouter.

When we have the `async` keyword in place then errors thrown by `getKey()` will appear as a rejecting promise thanks to the `Promise.all()` for getting all resources. However with this chance errors thrown by `getKey()` will just appear as a hard error.

So leaving this change in place would mean that getting all resources would need to cater for both rejecting promise and hard errors which is a bit unfortunate.

liamqma (approved):

PR #161, opened 2022-08-09
Fix issue with jsdom no longer adding window.name

timse:
jsdom no longer sets window.name - while this script doesnt entirely rely on that the whole detection can still be simplified


theKashey:
Wondering if https://github.com/iliakan/detect-node can be used for the same purpose. And it doesn't - then this sort of configuration should be left up to the particular user.

PR #163, opened 2022-09-07
Add compatibility for goBack and goForward with history@5

danskyr:
In `history@5` the `goForward` and `goBack` functions were renamed to `forward` and `back`. This means that `react-resource-router` will currently fail on these actions when used with `history@5`.

This PR fixes that and adds compatibility for these two actions with both `history@4` and `history@5`.

### DIFF
--- a/src/controllers/router-store/index.tsx
+++ b/src/controllers/router-store/index.tsx
@@ -309,6 +309,7 @@ const actions: AllRouterActions = {
     ({ getState }) => {
       const { history } = getState();
 
+      // history@4 uses goBack(), history@5 uses back()
       if ('goBack' in history) {
         history.goBack();
       } else if ('back' in history) {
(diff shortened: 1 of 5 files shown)


liamqma (approved):

src/controllers/router-store/index.tsx
```diff
@@ -309,15 +309,27 @@ const actions: AllRouterActions = {
     ({ getState }) => {
       const { history } = getState();
 
-      history.goBack();
+      if ('goBack' in history) {
```
liamqma, line 312: can you add a comment indicating `goBack` is v4 and `back` is v5 please?

liamqma (approved):

gorakong:
hey @liamqma and @albertogasparin, feel free to disregard the CLA check and merge the PR (if it's ready) since we've verified that these commits all come from an Atlassian 🙂.

PR #169, opened 2022-11-01
Modify prefetch delay and support focus and pointer events to prefetch

dalehurwitz:
**Changes**:

1. Modifies the prefetch delay from `300` -> `225`. This new number is based on manual testing and is a best-guess.
2. Extends `prefetch="hover"` to fire prefetch on `focus` and `pointerDown` events. This improves the prefetch experience:
- On `pointerDown` the prefetch will be fired immediately rather than waiting for the setTimeout to complete
- Browser tests (e.g. Cypress, Selenium) will be able to fire the prefetch when using their respective `.click` methods

### DIFF
--- a/src/common/utils/use-timeout/index.ts
+++ b/src/common/utils/use-timeout/index.ts
@@ -1,0 +1,22 @@
+import { useCallback, useRef } from 'react';
+
+export const useTimeout = (delay: number) => {
+  const timeoutId = useRef<NodeJS.Timeout>();
+
+  const schedule = useCallback(
+    (callback: () => void) => {
+      if (timeoutId.current) {
+        clearTimeout(timeoutId.current);
+      }
+
+      timeoutId.current = setTimeout(callback, delay);
+    },
+    [delay]
+  );
+
+  const cancel = useCallback(() => {
+    clearTimeout(timeoutId.current);
+  }, []);
+
+  return { schedule, cancel };
+};
--- a/src/ui/link/index.tsx
+++ b/src/ui/link/index.tsx
@@ -8,6 +8,7 @@ import {
   useState,
   MouseEvent,
   KeyboardEvent,
+  FocusEvent,
 } from 'react';
 
 import { LinkProps, Route } from '../../common/types';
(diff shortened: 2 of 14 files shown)


src/ui/link/index.tsx
```diff
@@ -83,13 +87,24 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [route, linkDestination, routerActions]);
 
+    const schedulePrefetch = useCallback(() => {
+      prefetchRef.current = setTimeout(triggerPrefetch, PREFETCH_DELAY);
```
albertogasparin, line 91: Should we check that `current` is not set yet? I wonder if there is a risk that `schedulePrefetch` is called twice and so we end up with 2 timers being set, but only one getting cancelled 🤔

src/ui/link/index.tsx
```diff
@@ -83,13 +87,24 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [route, linkDestination, routerActions]);
 
+    const schedulePrefetch = useCallback(() => {
```
albertogasparin, line 90: Nit: Wonder if we could optimise this a bit by having a smarter `prefetchRef` object instead of multiple `useCallback`:
```
const prefetchRef = useRef({
  timer: null,
  schedule() { ... },
  cancel() { ... },
})
```

albertogasparin (approved):

src/common/utils/use-timeout/index.ts
```diff
@@ -0,0 +1,22 @@
...
+      if (timeoutId.current) {
+        clearTimeout(timeoutId.current);
+      }
+
+      timeoutId.current = setTimeout(callback, delay);
+    },
+    [delay]
+  );
+
+  const cancel = useCallback(() => {
```
liamqma, line 17: What's the purpose of `useCallback` here? Is it to ensure `clearTimeout(timeoutId.current)` runs once only regardless how many times `cancel` get called?

liamqma (approved):

PR #185, opened 2023-01-09
Refactor components

MonicaOlejniczak:
## Changes
* Colocate tests, alphabetise imports and exports in affected files
* Flatten `subscribers/*` into `resource-subscriber` and `router-subscriber` for improved visibility
* Remove unused `RouterSubscriber` and `RouterActionsSubscriber` in the `router-store` from source
* Remove extra dot in tsconfig exclude
* Rename `useRouterStoreStatic` to match existing `useRouterStoreActions` naming convention and update usages
* Update following implementations to prefer:
  * `useRouter` over `RouterSubscriber` in `Redirect` component
  * `useRouterActions` over `RouterActionsSubscriber ` in `RouterActions` component
  * `useRouterStore` over `RouterSubscriber` from the router-store in `RouterSubscriber` component
  * `useRouterStore` over `RouterSubscriber` in `withRouter`
  * `useRouter` over `RouteComponent` in `RouteComponent`


src/controllers/router-subscriber/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { RouterActionsType, RouterState } from '../router-store/types';
+
+export type RouterSubscriberProps = {
+  children: (state: RouterState, actions: RouterActionsType) => ReactNode;
+};
+
+export const RouterSubscriber = ({ children }: RouterSubscriberProps) => {
+  const [state, { listen, ...actions }] = useRouterStore();
+
+  if (!state.unlisten && !state.isStatic) {
```
MonicaOlejniczak, line 13: @albertogasparin Do you know if it's safe to remove this condition? I think it should be expected that the `Router` container is initialised by consumers.

albertogasparin: The truth is... I'm no longer sure how legacy pages work in this context. As we now have a router there too, but (if I recall correctly) this was meant to fix cases where we did not fully initialise. I'd recommend to keep as is (to avoid issue with the well overdue refactoring that you are doing) and maybe re-evaluate with a patch directly in Jira to see if something breaks

albertogasparin (approved):

PR #186, opened 2023-01-10
Restructure hooks

MonicaOlejniczak:
## Changes
* Extract hooks from `controllers/hooks/router-store` into separate files
* Make remaining test names consistent
* Move hooks in `controllers/hooks` to `controllers` which will become less unwieldy when the routers are flattened
* Rename `__tests__/integration/test.tsx` to `__tests__/integration.test.tsx`


src/controllers/use-router-actions/index.ts
```diff
@@ -0,0 +1,11 @@
+import { useRouterStoreActions } from '../router-store';
+import { RouterActionsType } from '../router-store/types';
+
+/**
+ * Hook for accessing the public router store without re-rendering on route change
+ */
+export const useRouterActions = (): RouterActionsType => {
```
albertogasparin, line 7: FYI RSS now supports this pattern out of the box. We could kill this and change `useRouterStoreActions` to be:
const useRouterActions = createActionsHook(Store)

albertogasparin (approved):

src/controllers/use-router-actions/index.ts
MonicaOlejniczak: Thanks for the info, I'll do that in another pull request for all of them

PR #191, opened 2023-01-17
Refactor router store tests

MonicaOlejniczak:
These changes refactors the router store tests, where some tests in the router store are moved to integration tests and vice versa, ensuring that the test suite is better separated as unit vs integration tests.

## Changes
* Move `sends the expected path to history` integration tests to router store
* Move `history listen resource store interop` route transition tests from the router store to integration tests, and consolidate them into two tests rather than four
* Rename describe blocks and test cases as appropriate
* Test the `RouterContainer` over `Router` in the router store tests
* Test history v4/v5 for each test in router store, rather than a few
* Use `getRouterState()` or a local `getState` over `RouterStoreSubscriber` to cleanup the tests
* Use common render function in router store tests

### DIFF
--- a/src/__tests__/integration.test.tsx
+++ b/src/__tests__/integration.test.tsx
@@ -1,16 +1,17 @@
 import { mount } from 'enzyme';
 import { createMemoryHistory } from 'history';
-import React from 'react';
+import React, { Fragment } from 'react';
 import { defaultRegistry } from 'react-sweet-state';
 
 import { isServerEnvironment } from '../common/utils/is-server-environment';
 import { createResource, ResourceStore } from '../controllers/resource-store';
 import {
+  Route,
   RouteComponent,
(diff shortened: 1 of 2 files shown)


liamqma (approved):

src/__tests__/integration.test.tsx
```diff
@@ -1,16 +1,17 @@
 import { mount } from 'enzyme';
```
albertogasparin, line 1: Would had been nice to move to RTL, but I guess that is a bigger task

albertogasparin (approved):

src/__tests__/integration.test.tsx
MonicaOlejniczak: Would be nice! It should be easier to do with these refactors though

PR #192, opened 2023-01-17
Introduce plugins and extract resources as a plugin

bobrovnik:
That's just an API preview, so we can discuss it further


src/controllers/loader/index.ts
```diff
@@ -0,0 +1,23 @@
...
+export const createCombinedLoader = (props: Props) => {
+  const { context, resourceData, isStatic } = props;
+  const entryPoint = entryPointsLoader();
+  const resources = resourcesLoader({
+    context,
+    resourceData,
+    isStatic,
+  });
+
+  return combine(entryPoint, resources);
```
albertogasparin, line 22: An alternative to this combine API (very redux-like) could be to just pass an array of loaders. I find it easier to digest imo.
It also means that Props are specific to each individual loader, instead of having `resourceData` passed even when entryPoint does not care about it

src/controllers/router-store/index.tsx
```diff
@@ -398,6 +443,33 @@ const actions: AllRouterActions = {
         history[updateType](updatedRelativePath);
       }
     },
+  loadRoute:
```
albertogasparin, line 446: This action name is a bit confusing as might be interpreted as loading the full route (while instead we just load the loader 😅 . So `loadLoader` is not great either, but maybe `runLoaderLoad` ? 🤔

MonicaOlejniczak: I suggested the `loadRoute` name because in the case of EntryPoints (or any custom loader) you load both the component and data. It also does not have to be limited to this, it could load other things within the route (e.g. other components related to the page)

src/controllers/loader/index.ts
MonicaOlejniczak: Yeah, that should be fine, as it hides the `combine` as an implementation detail tbh

I'm not sure what you mean about the `resourceData` though, that should be the case with this API anyway.

src/common/types.ts
```diff
@@ -335,3 +335,23 @@ export type FindRouterContextOptions = {
   location: Location;
   basePath?: string;
 };
+
+export type LoaderAPI<T = any> = {
+  onBeforeRouteChange: (params: {
```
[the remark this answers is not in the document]
MonicaOlejniczak: Agreed

src/common/types.ts
```diff
@@ -335,3 +335,24 @@ export type FindRouterContextOptions = {
...
+  }) => void;
+  load: (
+    context: RouterContext & {
+      prevLocationContext?: RouterContext;
+    }
+  ) => T;
+  prefetch: (context: RouterContext) => void;
+};
+
+export type Loader<T> = (props: {
```
MonicaOlejniczak, line 353: Name this `ResourceLoader` since these arguments are specific to resources

src/controllers/router-store/index.tsx
```diff
@@ -116,8 +131,13 @@ const actions: AllRouterActions = {
         basePath,
         location: history.location,
         action: history.action,
+        loader,
       });
-      getResourceStore().actions.hydrate({ resourceContext, resourceData });
+
+      if (loader.hydrate) {
```
MonicaOlejniczak, line 137: If you rebase the latest changes from master, this method should be gone.

src/common/types.ts
```diff
@@ -335,3 +335,24 @@ export type FindRouterContextOptions = {
   location: Location;
   basePath?: string;
 };
+
+export type LoaderAPI<T = Record<string, unknown>> = {
```
MonicaOlejniczak, line 339: I would just call it `Loader`

src/controllers/loader/index.ts
```diff
@@ -0,0 +1,23 @@
...
+import { resourcesLoader } from '../../resources/loader';
+
+type Props = {
+  context: ResourceStoreContext | undefined;
+  resourceData: any;
+  timeout?: number;
+  isStatic?: boolean;
+};
+
+export const createCombinedLoader = (props: Props) => {
```
MonicaOlejniczak, line 13: Was this intended to be part of the public API? Because imo entrypoints shouldn't be in here. I was thinking that the `Router` can take in a `loader` or `loaders` (if you follow Alberto's suggestion) that defaults to the resource loader, making it a non-breaking change. Then consumers that want to combine them just opt into it through the new API. That way you separate the `Props` from `createCombinedLoader` into a `createResourceLoader`

src/controllers/router-store/index.tsx
```diff
@@ -83,8 +91,14 @@ const actions: AllRouterActions = {
         routes,
         location: history.location,
         action: history.action,
+        loader,
       });
-      getResourceStore().actions.hydrate({ resourceContext, resourceData });
+
+      if (loader.hydrate) {
```
MonicaOlejniczak, line 97: It should still be possible to put this into the resource loader constructor time, and I don't think it's a big deal if you have to construct two separate loaders (one for the server, one for the client)

src/__tests__/integration.test.tsx
```diff
@@ -226,7 +226,7 @@ describe('<Router /> client-side integration tests', () => {
       };
     }
 
-    it('previous data when transitioning to the same route and resource keys', async () => {
+    it.only('previous data when transitioning to the same route and resource keys', async () => {
```
albertogasparin, line 229: ⚠️

src/controllers/router-store/index.tsx
```diff
@@ -366,6 +345,33 @@ const actions: AllRouterActions = {
...
+        nextContext = findRouterContext(routes, { location, basePath });
+      }
+
+      if (nextContext == null) return;
+      const nextLocationContext = nextContext;
+
+      batch(() => {
+        loader.prefetch(nextLocationContext);
+        // prefetchResources(nextResources, nextLocationContext, {});
+        if (onPrefetch) onPrefetch(nextLocationContext);
```
albertogasparin, line 371: Now I wonder: should this be moved to a loader too? This could very well be a `webpackLoader` and could handle what `onPrefetch` does in Jira without the need of a specific prop 👀

src/common/utils/combine/index.ts
```diff
@@ -0,0 +1,35 @@
...
+    beforeLoad: params => {
+      loaders.forEach(loader => {
+        if (loader.beforeLoad !== undefined) {
+          loader.beforeLoad(params);
+        }
+      });
+    },
+
+    load: loadParams => {
+      return loaders.reduce(
```
albertogasparin, line 21: Can you explain to me why this is reduce and the others are foreach?

src/controllers/router/index.tsx
```diff
@@ -59,16 +75,35 @@ Router.requestResources = async ({
...
   bootstrapStore({
     ...bootstrapProps,
     history: history || createMemoryHistory({ initialEntries: [location] }),
+    loader,
   });
 
-  await requestRouteResources({ timeout });
+  // await requestRouteResources({ timeout });
+
+  await loadRoute().resources;
```
albertogasparin, line 106: What if I (as a consumer) have another loader that needs to be awaited?
Isn't it better if this is:
```js
await Promise.all(loadRoute())
```
And `loadRoute` is just a map of the calls of all loader `.load()` calls?

src/entry-points/use-entry-point/index.ts
```diff
@@ -0,0 +1,11 @@
+import { useRouter } from '../../controllers/use-router';
+
+export const useEntryPoint = (): any => {
+  const [{ route }] = useRouter();
+
+  if (!route.entryPoint) {
+    return undefined;
+  }
+
+  return { entryPointReference: route?.entryPoint }; // loadEntryPoint(...)
```
albertogasparin, line 10: `route` is optionally null here but `!route.entryPoint` does not check for that

src/common/types.ts
```diff
@@ -326,3 +326,23 @@ export type FindRouterContextOptions = {
...
+    prevContext: RouterContext;
+    nextContext: RouterContext;
+  }) => void;
+  load: (context: RouterContext, prevContext?: RouterContext) => T;
+  prefetch: (context: RouterContext) => void;
+};
+
+export type CombinedLoader<T = Record<string, any>> = {
+  hydrate: () => void;
+  beforeLoad: (params: {
```
albertogasparin, line 342: Should this follow the same shape as `load` ? 
```
beforeLoad: (prevContext: RouterContext, nextContext: RouterContext ) => void;
```

src/common/types.ts
```diff
@@ -326,3 +326,23 @@ export type FindRouterContextOptions = {
...
+    prevContext: RouterContext;
+    nextContext: RouterContext;
+  }) => void;
+  load: (context: RouterContext, prevContext?: RouterContext) => T;
+  prefetch: (context: RouterContext) => void;
```
albertogasparin, line 347: Is there any value on providing `prevContext` here too? 🤔

src/common/types.ts
MonicaOlejniczak: Yeah, or the other way around is more flexible for future changes.

`beforeLoad({ nextContext, prevContext })`

`Load({ nextContext, prevContext })`

src/controllers/router-store/index.tsx
MonicaOlejniczak: Yeah, I think it can too. Jira can easily add a custom loader to maintain parity behaviour. It would be a breaking change though but a minor one at that since there wouldn't be too many consumers.

src/common/types.ts
```diff
@@ -326,3 +326,23 @@ export type FindRouterContextOptions = {
...
+  hydrate?: () => void;
+  beforeLoad?: (params: {
+    prevContext: RouterContext;
+    nextContext: RouterContext;
+  }) => void;
+  load: (context: RouterContext, prevContext?: RouterContext) => T;
+  prefetch: (context: RouterContext) => void;
+};
+
+export type CombinedLoader<T = Record<string, any>> = {
```
MonicaOlejniczak, line 340: Should be removed in favour of Loader since they're now the same

src/controllers/router-store/test.tsx
```diff
@@ -142,6 +165,7 @@ describe('RouterStore', () => {
           route: routes[0],
           routes: routes,
           unlisten: expect.any(Function),
+          loader: expect.any(Object),
```
MonicaOlejniczak, line 168: Should just be `loader,`

src/controllers/router/index.tsx
```diff
@@ -24,6 +26,7 @@ export const Router = ({
   resourceContext,
   resourceData,
   routes,
+  loaders,
```
MonicaOlejniczak, line 29: /nit alphabetise props

src/controllers/router/index.tsx
```diff
@@ -33,6 +36,18 @@ export const Router = ({
     };
   }, []);
 
+  const loader = useMemo(() => {
+    if (loaders) return combine(...loaders);
```
MonicaOlejniczak, line 40: Shouldn't need to spread `...loaders`

src/controllers/router/index.tsx
```diff
@@ -33,6 +36,18 @@ export const Router = ({
...
+
+    // default 'loaders' fallback for the first relase
+    const resources = createResourcesLoader({
+      context: resourceContext,
+      resourceData,
+    });
+
+    return combine(resources);
```
MonicaOlejniczak, line 48: `return resources`

.eslintrc.js
```diff
@@ -47,6 +47,7 @@ module.exports = {
     'react-hooks/exhaustive-deps': 'warn',
     'react-hooks/rules-of-hooks': 'error',
     semi: 'off',
+    'react/jsx-sort-props': 'error',
```
MonicaOlejniczak, line 50: Can you make this a separate pull request?

src/common/utils/combine/index.ts
```diff
@@ -0,0 +1,39 @@
+import type { Plugin } from '../../types';
+
+export const combine = (plugins: Plugin[]): Required<Plugin> => {
+  return {
+    onHydrate: () => {
+      plugins.forEach(plugin => {
+        if (plugin.onHydrate !== undefined) {
```
albertogasparin, line 7: I'd write these as:
```
if (plugin.onHydrate != null) {
```
Just to be more gentle with somebody setting the method to `null`

albertogasparin: Also another way to specify this "combination" is to have an API like:
```
plugins.run('onHydrate')
```
That would allow us to expand the supported methods without copy pasting these lines for all methods.

MonicaOlejniczak (changes requested):

src/controllers/router/index.tsx
```diff
@@ -52,27 +69,57 @@ export const Router = ({
...
 
   return getResourceStore().actions.getSafeData();
 };
 
+Router.loadRoute = ({ history, plugins, routes }: LoadRouteParams) => {
+  const { bootstrapStore, loadRoute } = getRouterStore().actions;
+
+  bootstrapStore({
+    routes,
+    history: history,
```
MonicaOlejniczak, line 116: /nit `history,`

src/resources/utils/get-resources-safe-data/index.ts
```diff
@@ -0,0 +1,5 @@
+import { getResourceStore } from '../../../controllers/resource-store/index';
+
+export const getResourcesSafeData = () => {
```
MonicaOlejniczak, line 3: Can we rename this? Maybe `getSerializedResources`?

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "react-resource-router",
-  "version": "0.21.2",
+  "version": "0.21.3-dev-192.0.0.0",
```
MonicaOlejniczak, line 3: You should revert this

src/controllers/router/index.tsx
MonicaOlejniczak: Seems this is handled internally now

MonicaOlejniczak (approved):

examples/hydration-with-plugins/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+    resourceData: data,
+  });
+
+  const App = () => {
+    return (
+      <Router
+        basePath="/hydration-with-plugins"
+        history={myHistory}
+        plugins={[resourcesPlugin]}
+        resourceData={data}
```
albertogasparin, line 53: This should no longer be here in the plugin example, right?

src/common/types.ts
```diff
@@ -326,3 +326,16 @@ export type FindRouterContextOptions = {
   location: Location;
   basePath?: string;
 };
+
+export type Plugin<T = Record<string, any>> = {
+  onHydrate?: () => void;
```
albertogasparin, line 331: Should we pass the `context` here too? We do have it available...

src/resources/plugin/index.ts
```diff
@@ -0,0 +1,100 @@
...
+            route,
+            match,
+            query,
+          },
+          { timeout }
+        ),
+      };
+    },
+    onRoutePrefetch: (context: RouterContext) => {
+      const { route, match, query } = getRouterState();
```
albertogasparin, line 85: Isn't this data available on `context: RouterContext` ? `getRouterState` is a sort of private API, so if possible I'd like plugins to not reach into it

src/controllers/router-store/index.tsx
```diff
@@ -102,6 +109,7 @@ const actions: AllRouterActions = {
       const { route, match, query } = getState();
```
albertogasparin, line 109: What is the future of `requestRouteResources` action? Is this superseded by one of the plugin methods? Can it be deleted?

src/controllers/router-store/index.tsx
```diff
@@ -366,6 +342,32 @@ const actions: AllRouterActions = {
...
+
+      if (nextContext == null) return;
+      const nextLocationContext = nextContext;
+
+      batch(() => {
+        plugins.onRoutePrefetch(nextLocationContext);
+        if (onPrefetch) onPrefetch(nextLocationContext);
+      });
+    },
+  // ...
```
albertogasparin, line 370: Nit: remove

src/controllers/router-store/test.tsx
```diff
@@ -49,10 +66,18 @@ describe('RouterStore', () => {
...
+        createPlugins({
+          context: props.resourceContext,
+          resourceData: props.resourceData,
+        });
+
       mount(
         <RouterContainer
           history={history}
           isGlobal
+          plugins={plugins}
```
albertogasparin, line 80: The fact that we have a `plugins` API that is `[plugin, plugin]` and this `plugins` that is the result of `combine([plugin, plugin])` caused me few headaches while reviewing. My brain was expecting `plugins` props to be the same, not change in some places but not in others.

So what about 
- we use `combine` only inside the bootstrap action. So the props/args called `plugins` are always arrays of plugin instances. 
- We can pick a different name for the value in the router store (eg `combinedPlugins`) so it is more obvious that it is no longer the array of plugins.

The result is also that these tests don't need to know about `combine` at all

src/controllers/router-store/types.ts
```diff
@@ -34,6 +35,7 @@ type PrivateStateProperties = {
   onPrefetch?: (routerContext: RouterContext) => void;
   routes: Routes;
   unlisten: UnlistenHistory | null;
+  plugins: Required<Plugin>;
```
albertogasparin, line 38: Let's call this `combinedPlugins` or something similar so it is more clear that is is the "unified" API and not an array.

src/controllers/router-store/types.ts
```diff
@@ -47,6 +49,7 @@ export type ContainerProps = {
   resourceContext?: ResourceStoreContext;
   resourceData?: ResourceStoreData;
   routes: Routes;
+  plugins: Required<Plugin>;
```
albertogasparin, line 52: I'd prefer this to be `Required<Plugin[]>`

albertogasparin (approved):

MonicaOlejniczak (dismissed):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-resource-router",
-  "version": "0.21.2",
```
albertogasparin, line 3: This should be reverted

albertogasparin (approved):

src/controllers/router-store/test.tsx
```diff
@@ -172,6 +200,32 @@ describe('RouterStore', () => {
...
+        const plugins = [plugin];
+
+        renderRouterContainer({
+          plugins,
+        });
+
+        expect(plugin.hydrate).toBeCalled();
+      });
+
+      it('plugin onRouteLoad is called on initial render', () => {
```
MonicaOlejniczak, line 216: `routeLoad`

src/controllers/router-store/test.tsx
```diff
@@ -172,6 +200,32 @@ describe('RouterStore', () => {
         });
       });
 
+      it('onHydrate for each plugin is called', () => {
```
MonicaOlejniczak, line 203: `hydrate`

src/common/types.ts
```diff
@@ -326,3 +326,20 @@ export type FindRouterContextOptions = {
   location: Location;
   basePath?: string;
 };
+
+export type Plugin = {
+  hydrate?: (params: { context: RouterContext }) => void;
```
MonicaOlejniczak, line 331: Do we still need this?

src/controllers/router/index.tsx
```diff
@@ -52,25 +64,31 @@ export const Router = ({
...
  * The entry point for requesting resource data on the server.
  * Pass the result data into the router as a prop in order to hydrate it.
  */
 Router.requestResources = async ({
   location,
   history,
   timeout,
-  ...bootstrapProps
+  routes,
+  resourceContext: context,
```
MonicaOlejniczak, line 76: You should add `basePath` and pass it to `invokePluginLoad`

src/controllers/router-store/index.tsx
```diff
@@ -82,9 +81,10 @@ const actions: AllRouterActions = {
         routes,
         location: history.location,
         action: history.action,
+        plugins,
       });
 
-      getResourceStore().actions.hydrate({ resourceContext, resourceData });
+      plugins.forEach(p => p.hydrate?.({ context: routerContext }));
```
MonicaOlejniczak, line 87: Remove this one

MonicaOlejniczak (approved):

PR #193, opened 2023-01-18
Consolidate routers

MonicaOlejniczak:
A continutation of https://github.com/atlassian-labs/react-resource-router/pull/145 with the following additions:
* Make `history` prop required on `Router` as it being optional previously created no reason to use `MemoryRouter` for testing
* Update SSR example to use `Router` with a memory history passed into it
* Add hash router example to reproduce https://github.com/atlassian-labs/react-resource-router/issues/111
* Refactor router tests
* Remove unused `createLocation` function
  * Used in both [StaticRouter](https://github.com/atlassian-labs/react-resource-router/blob/master/src/controllers/static-router/index.tsx) and [UniversalRouter](https://github.com/atlassian-labs/react-resource-router/blob/master/src/controllers/universal-router/index.tsx) to pass `location` into `bootstrapStore` and `bootstrapStoreUniversal` respectively
  * `location` is always set to `history.location` and thus goes unused in [bootstrapStore](https://github.com/atlassian-labs/react-resource-router/blob/master/src/controllers/router-store/index.tsx#L74-L86) and [bootstrapStoreUniversal](https://github.com/atlassian-labs/react-resource-router/blob/master/src/controllers/router-store/index.tsx#L108-L119)
  * Remove `props.location` from `Router` and prefer `history`


src/controllers/router-store/index.tsx
```diff
@@ -63,65 +61,32 @@ const actions: AllRouterActions = {
...
-        basePath,
-        location: history.location,
-        action: history.action,
-      });
       getResourceStore().actions.hydrate({ resourceContext, resourceData });
 
-      if (!isServerEnvironment()) {
-        dispatch(actions.listen());
-      }
+      dispatch(actions.listen());
```
albertogasparin, line 89: No need to prevent this from being called on server?

src/controllers/router/index.tsx
```diff
@@ -1,77 +1,89 @@
...
+  location,
+  history,
+  timeout,
+  ...bootstrapProps
+}: RequestResourcesParams) => {
+  const { bootstrapStore, requestRouteResources } = getRouterStore().actions;
+
+  bootstrapStore({
+    ...bootstrapProps,
+    history: history || createMemoryHistory({ initialEntries: [location] }),
```
albertogasparin, line 68: If `history` is now required as prop, should we mandate it here too? Wonder if this might cause problem if we pass a history v4 and consumer is using history v5

src/controllers/router/index.tsx
```diff
@@ -1,77 +1,89 @@
...
-      onPrefetch,
-    } = this.props;
+    return () => {
+      unlisten && unlisten();
+    };
+  }, []);
 
-    return (
+  return (
+    <ResourceContainer isGlobal={isGlobal}>
```
albertogasparin, line 37: Based on our earlier discussion, wonder if `isGlobal` here is really doing what we think. As if `false`, would `getResourceStore()` actually work??

src/controllers/router-store/index.tsx
MonicaOlejniczak: Ah I think this was ported incorrectly from your original changes 😂

src/controllers/router/index.tsx
MonicaOlejniczak: It's a peer dependency, so it shouldn't matter. But I can change it if you like

albertogasparin: We said we wanna deprecate this API later on anyway, so fine to leave as is

src/controllers/router/index.tsx
MonicaOlejniczak: Well the integration tests do fail when `isGlobal` is false on the `ResourceContainer` and that is using the public API (i.e. nothing is mocked). Shall I hardcode the resource container `isGlobal` to true?

src/controllers/router-store/index.tsx
MonicaOlejniczak: Should be fixed now

albertogasparin (approved):

PR #197, opened 2023-04-21
Ensure plugin load methods are called only when path/params/query change

bobrovnik:
Changing when plugin.routeLoad and plugin.beforeRouteLoad is happening. 

The current behavior is to load Route anytime URL changes. I'm changing it to only when Route path/params/query change.

"Resources" plugin is an exception, as it currently relies on old behavior.


src/common/utils/is-same-route/index.ts
```diff
@@ -0,0 +1,30 @@
+import { Match } from '../../../common/types';
+
+type ObjectToCompare = {
+  [key: string]: string | null | typeof undefined;
+};
+const isDeepEqual = (object1: ObjectToCompare, object2: ObjectToCompare) => {
```
albertogasparin, line 6: Nit: this is in reality `isShallowEqual`

src/controllers/router-store/index.tsx
```diff
@@ -149,28 +150,46 @@ const actions: AllRouterActions = {
...
 
           batch(() => {
-            plugins.forEach(p =>
-              p.beforeRouteLoad?.({
-                context: prevContext,
-                nextContext,
-              })
-            );
+            plugins.forEach(p => {
+              if (p.id === 'resources-plugin') {
```
albertogasparin, line 165: Can we combine the ifs?
```
if (p.id === 'resources-plugin' || !sameRoute) { ... }
```

src/common/utils/is-same-route/index.ts
```diff
@@ -0,0 +1,30 @@
+import { Match } from '../../../common/types';
+
+type ObjectToCompare = {
```
MonicaOlejniczak, line 3:
```suggestion
type ObjectToCompare = Record<string, string | null | undefined>;
```

src/common/utils/is-same-route/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  [key: string]: string | null | typeof undefined;
+};
+const isShallowEqual = (object1: ObjectToCompare, object2: ObjectToCompare) => {
+  const objKeys1 = Object.keys(object1);
+  const objKeys2 = Object.keys(object2);
+
+  if (objKeys1.length !== objKeys2.length) return false;
+
+  for (const key of objKeys1) {
```
MonicaOlejniczak, line 12:
```suggestion
return objKeys1.every((key) => object1[key] === object2[key]);
```

src/common/utils/is-same-route/index.ts
```diff
@@ -0,0 +1,30 @@
+import { Match } from '../../../common/types';
+
+type ObjectToCompare = {
+  [key: string]: string | null | typeof undefined;
+};
+const isShallowEqual = (object1: ObjectToCompare, object2: ObjectToCompare) => {
```
MonicaOlejniczak, line 6: Could just shorten it to `o1` and `o2`

at-nathan (approved):

albertogasparin (approved):

PR #198, opened 2023-04-24
Fix exported Resources plugin types

bobrovnik:
Fixing types export mismatch between `./package.json` (exports "cjs") and `./resources/package.json` (exports "esm")

### DIFF
--- a/resources/package.json
+++ b/resources/package.json
@@ -1,5 +1,5 @@
 {
-  "main": "../build/cjs/resources.js",
-  "module": "../build/esm/resources.js",
-  "types": "../build/cjs/resources.d.ts"
+  "main": "../build/cjs/resources/index.js",
+  "module": "../build/esm/resources/index.js",
+  "types": "../build/cjs/resources/index.d.ts"
 }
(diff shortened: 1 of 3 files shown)


pancaspe87 (dismissed):

resources/package.json
```diff
@@ -1,5 +1,5 @@
 {
   "main": "../build/cjs/resources.js",
   "module": "../build/esm/resources.js",
-  "types": "../build/esm/resources.d.ts"
```
albertogasparin, line 4: 🤔 Don't think this is the problem. From what I can see there is no `resources.x` at all, in any of the build folders

albertogasparin: Ok, was expecting an index inside resources folder. Instead we have both a `resources.js` and a `resources` folder. It might cause some headaches in future...

Anyway seems like we export types from `cjs` folder in the root level, so let's keep it consistent

albertogasparin (dismissed):

at-nathan (approved):

albertogasparin (approved):

MonicaOlejniczak (approved):

PR #202, opened 2023-05-05
Use public NPM registry instead of Atlassian one

bobrovnik:
Looks like we've private NPM registry https://packages.atlassian.com/api/npm/npm-remote was used, which is a problem for non-Atlassian consumers


at-nathan (dismissed):

pancaspe87 (approved):

MonicaOlejniczak (commented):
Can you add an [.npmrc](https://docs.npmjs.com/cli/v9/configuring-npm/npmrc) file that uses the public registry, so this doesn't happen again?

at-nathan (approved):

PR #203, opened 2023-05-08
[EARLY FEEDBACK] Add option to avoid Plugins reload when staying within same route

bobrovnik:
Getting PR early to get feedback on naming and general approach.

Inspired by https://reactrouter.com/en/main/route/should-revalidate

### DIFF
--- a/src/common/utils/should-reload-when-route-match-changes/index.ts
+++ b/src/common/utils/should-reload-when-route-match-changes/index.ts
@@ -5,25 +5,12 @@ type Options = {
   params?: string[];
 };
 export const shouldReloadWhenRouteMatchChanges =
-  (options: Options) =>
+  ({ params = [], query = [] }: Options) =>
   ({ context, prevContext }: Parameters<ShouldReloadFunction>[0]) => {
-    if (
-      options?.query !== undefined &&
-      options?.query.some(
-        q => context.match.query[q] !== prevContext.match.query[q]
-      )
-    ) {
-      return true;
-    }
-
-    if (
-      options?.params !== undefined &&
-      options?.params.some(
+    return (
+      params.some(
         p => context.match.params[p] !== prevContext.match.params[p]
-      )
-    ) {
-      return true;
-    }
-
-    return false;
+      ) ||
+      query.some(q => context.match.query[q] !== prevContext.match.query[q])
+    );
   };
--- a/src/common/utils/should-reload-when-route-match-changes/index.ts
+++ b/src/common/utils/should-reload-when-route-match-changes/index.ts
@@ -1,0 +1,29 @@
+import { ShouldReloadFunction } from '../../types';
+
+type Options = {
+  query?: string[];
+  params?: string[];
+};
+export const shouldReloadWhenRouteMatchChanges =
+  (options: Options) =>
+  ({ context, prevContext }: Parameters<ShouldReloadFunction>[0]) => {
+    if (
+      options?.query !== undefined &&
+      options?.query.some(
+        q => context.match.query[q] !== prevContext.match.query[q]
+      )
+    ) {
+      return true;
+    }
+
+    if (
+      options?.params !== undefined &&
+      options?.params.some(
+        p => context.match.params[p] !== prevContext.match.params[p]
+      )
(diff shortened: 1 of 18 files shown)


albertogasparin (dismissed):

src/controllers/router-store/index.tsx
```diff
@@ -140,7 +141,15 @@ const actions: AllRouterActions = {
 
           batch(() => {
             plugins.forEach(p => {
-              if (p.id === 'resources-plugin' || !sameRoute) {
+              if (
+                shouldReload({
+                  context: nextContext,
+                  prevContext,
+                  pluginId: p.id,
+                  defaultShouldReload:
```
MonicaOlejniczak, line 149: Can `defaultShouldReload` and `sameRoute` just move to the internals of `shouldReload` considering it receives the `pluginId`?

src/controllers/router-store/utils/index.ts
```diff
@@ -64,3 +69,27 @@ export const updateQueryParams = (location: Location, query: Query): string => {
...
+}: Parameters<ShouldReloadFunction>[0]) => {
+  if (
+    context.route === prevContext.route &&
+    context.route.EXPERIMENTAL__shouldReload
+  ) {
+    const routeChoice = context.route.EXPERIMENTAL__shouldReload({
+      context,
+      prevContext,
+      pluginId,
+      defaultShouldReload,
```
MonicaOlejniczak, line 87: Why would a route need to care about the default value? It can be inferred by the `pluginId` as well, if needed

src/controllers/router-store/index.tsx
MonicaOlejniczak: You could avoid that if you pre-process the whole thing, but you would need one more iteration for each plugin (this is cheap, the shallow equal on the routes may be more expensive).

```
const shouldReloadByPlugin = new Map(
  plugins.map((plugin) =>
    [plugin.id, shouldReload({ context, prevContext, pluginId: plugin.id })]
  )
);

// example usage
for (const plugin of plugins) {
  if (shouldReloadByPlugin.get(plugin.id)) {}
}


```

src/controllers/router-store/index.tsx
```diff
@@ -298,7 +309,13 @@ const actions: AllRouterActions = {
       );
 
       if (updatedPath !== existingPath) {
+        if (options?.avoidRoutePluginsLoad === true)
```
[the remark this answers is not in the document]
MonicaOlejniczak: Do we need this behaviour at the moment?

src/common/utils/should-reload-when-route-match-changes/index.ts
```diff
@@ -0,0 +1,29 @@
+import { ShouldReloadFunction } from '../../types';
+
+type Options = {
+  query?: string[];
+  params?: string[];
+};
+export const shouldReloadWhenRouteMatchChanges =
+  (options: Options) =>
+  ({ context, prevContext }: Parameters<ShouldReloadFunction>[0]) => {
+    if (
```
MonicaOlejniczak, line 10: Should be equivalent to this:

```
export const shouldReloadWhenRouteMatchChanges =
  ({ params = [], query = [] }: Options = {}) =>
  ({ context, prevContext }: Parameters<ShouldReloadFunction>[0]) => {
    return
      params.some((p) => context.match.params[p] !== prevContext.match.params[p])
      || query.some((q) => context.match.query[q] !== prevContext.match.query[q]);
  };
```

src/controllers/router-store/index.tsx
bobrovnik: make sense, let me do it

src/common/utils/should-reload-when-route-match-changes/index.ts
bobrovnik: updated

src/controllers/router-store/index.tsx
bobrovnik: done

at-nathan (approved):

MonicaOlejniczak (approved):

PR #206, opened 2023-05-23
Move all resources exports under react-resource-router/resources

bobrovnik:
Deprecating all `resources` related exports from `react-resource-router` and moving them to `react-resource-router/resources`


src/controllers/router-store/index.tsx
```diff
@@ -400,7 +400,7 @@ export const RouteResourceEnabledSubscriber = createSubscriber<
   Actions,
   boolean
 >(RouterStore, {
-  selector: state => Boolean(state.route && state.route.resources),
+  selector: state => Boolean(state.route && (state.route as any).resources),
```
albertogasparin, line 403: Should this be moved too? It can probably be recreated via `createRouterSelector`  (if we don't want to import router store)

albertogasparin (approved):

src/controllers/router-store/index.tsx
bobrovnik: good point

src/resources/controllers/route-resources-enabled-subscriber/index.ts
```diff
@@ -0,0 +1,17 @@
+import { createSubscriber } from 'react-sweet-state';
+
+import {
+  RouterStore,
+  type EntireRouterState,
+  type AllRouterActions,
+} from '../../../index';
+import type { RouteWithResources } from '../../common/types';
+
+export const RouteResourceEnabledSubscriber = createSubscriber<
```
MonicaOlejniczak, line 10: This is fine for now, but if these were separate packages then you'd want to use `createRouterSelector` as that is part of the public API

MonicaOlejniczak (dismissed):

docs/README.md
```diff
@@ -26,7 +26,7 @@ React Resource Router was developed by Atlassian for [Jira](https://www.atlassia
 Resources describe and provide the data required for your route. This data is safely stored and accessed via the [`useResource`](api/hooks.md#use-resource) hook or [`ResourceSubscriber`](api/components.md#resourcesubscriber) component.
 
 ```js
-import { createResource } from 'react-resource-router';
```
bobrovnik, line 29: How does this get released? when I merge PR or when I release new npm version?

src/resources/controllers/route-resources-enabled-subscriber/index.ts
bobrovnik: Something like this?
```
export const RouteResourceEnabledSubscriber = createRouterSelector(state =>
  Boolean(state.route && (state.route as RouteWithResources).resources)
);
```

docs/README.md
MonicaOlejniczak: I think when you merge to master

src/resources/controllers/route-resources-enabled-subscriber/index.ts
MonicaOlejniczak: I would be happy to kill it, personally. It would look more like this though, which you could provide in the patch notes

```javascript
const useHasRouteResources = createRouterSelector(state =>
  Boolean(state.route && (state.route as RouteWithResources).resources)
);

export const RouteResourceEnabledSubscriber = ({ children }) => {
  const hasResources = useHasRouteResources();

  return children(hasResources);
};
```

bobrovnik: Let's kill it

docs/README.md
bobrovnik: @MonicaOlejniczak do you think we should update docs in another PR? to be sure it's not going to get public until the new version is released?

MonicaOlejniczak: Either is fine

MonicaOlejniczak (approved):

PR #207, opened 2023-05-24
Making Router "plugins" prop required

bobrovnik:
- making Router "plugins" prop required
- removing `resources` plugin as a fallback when `Router.plugins` are not set


examples/hydration-with-plugins/index.tsx
```diff
@@ -19,9 +19,7 @@ const myHistory = createBrowserHistory();
 const appRoutes = [homeRoute];
 
 const getStateFromServer = async () => {
```
MonicaOlejniczak, line 21: Can we rename this set of examples to `examples/hydration`, not `examples/hydration-with-plugins`

MonicaOlejniczak (changes requested):

examples/hydration-with-plugins/index.tsx
bobrovnik: done

docs/api/components.md
```diff
@@ -33,6 +36,7 @@ import { appRoutes } from './routing';
 | ----------------- | ------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
 | `routes`          | `Routes[]`                | Your application's routes                                                                                                                     |
 | `history`         | `History`                 | The history instance for the router, if omitted memory history will be used (optional but recommended)                                        |
+| `plugins`         | `Plugin[]`                | Plugin allows you to hook into Router API and extra login on route load/prefetch/etc                                       |
```
bobrovnik, line 39: what do you guys think, should we 
1) keep `plugins` optional and default to empty array `[]`. Side effect - it might be unclear that you need to now pass `resourcesPlugin` to the Router when upgrading to the new RRR version.
2) make `plugins` prop required

at-nathan: I would go with option 1 as Router can be used without plugins. When upgrading with option 2 it will break typing and it might also null error for those not using types.

> it might be unclear that you need to now pass resourcesPlugin

Docs and release log should help with this right? Or maybe default to `[resourcesPlugin]`?

bobrovnik: > Docs and release log should help with this right?

In theory, if folks read it at all 

> Or maybe default to [resourcesPlugin]?
That's the exact behavior I'm trying to remove, in order to fully decouple `resources-plugin` from the Router

bobrovnik: @MonicaOlejniczak what do you think about it?

albertogasparin: I'd prefer option 2, make plugins prop required. It will make the breaking change / new API obvious and clear, forcing devs to check the release notes. As long as we document how to fix it in there too, it should require only few line changes (per Router use tho).
It will also open the door to remove unnecessary router props (resourceContext, resourceData), while if we keep the implicit creation we still need to allow those.

MonicaOlejniczak: I also prefer option 2 for the reasons Alberto mentioned

MonicaOlejniczak (approved):

docs/api/components.md
```diff
@@ -33,6 +36,7 @@ import { appRoutes } from './routing';
...
 | `basePath`        | `string`                  | Base path string that will get prepended to all route paths (optional)                                                                        |
 | `initialRoute`    | `Route`                   | The route your application is initially showing, it's a performance optimisation to avoid route matching cost on initial render(optional)     |
 | `resourceContext` | `ResourceContext`         | Custom contextual data that will be provided to all your resources' `getKey` and `getData` methods (optional)                                 |
```
albertogasparin, line 42: If we go with option 2, should this be removed?

examples/hooks/index.tsx
```diff
@@ -53,7 +54,12 @@ const Title = () => {
 
 const App = () => {
   return (
-    <Router basePath="/hooks" history={myHistory} routes={appRoutes}>
+    <Router
+      basePath="/hooks"
+      history={myHistory}
+      routes={appRoutes}
+      plugins={[createResourcesPlugin({})]}
```
albertogasparin, line 61: Is this a bad pattern? Would this create a new plugin instance on every re-render?

docs/api/components.md
bobrovnik: sounds good, thanks for your feedback team

examples/hooks/index.tsx
bobrovnik: good point

docs/api/components.md
bobrovnik: oh, I'm still working on docs. I'll push change shortly

examples/hooks/index.tsx
bobrovnik: Fixed

docs/api/components.md
bobrovnik: Done, it's ready for review

bobrovnik:
Before release, I still need to prepare a page about plugins. But I'll do it in another PR

albertogasparin (approved):

PR #209, opened 2023-06-13
Updates codeshift/cli to use current config spec

danieldelcore:
Updates codeshift.config to apply the latest config specification. 

- Use `require` instead of `require.resolve`
- Use `.ts` instead of `.js` since the codemods are not bundled (and so we lean on our internal ts parsing)
- Bumps dependencies


bobrovnik (approved):
@danieldelcore thanks for bumping the codeshift and updating config

danieldelcore:
Thanks @bobrovnik! Is there anything else I need to do before we merge this one :) ?

MonicaOlejniczak (approved):

PR #215, opened 2023-09-05
Added state to .push

AlexanderKaran:


src/controllers/router-store/index.tsx
```diff
@@ -157,6 +157,7 @@ const actions: AllRouterActions = {
               ...nextContext,
               location,
               action,
+              state: location.state,
```
albertogasparin, line 160: Given we are already storing `location`, does it make sense to store `state` here too? Wouldn't it be duplicated?

albertogasparin:
Great start! Would you mind:
- add `state` support also to the other imperative apis `.replace`, `pushTo`, `replaceTo` ?
- add `state` support to `Link` via props `<Link href="..." state={...} />` 
- ensure that it is available and returned in the hooks (eg in `useRouter`) by adding 1/2 tests there too

Thanks!!

src/controllers/router-store/index.tsx
AlexanderKaran: Good point. I can remove this line. I can either leave it inside the location or abstract it out and so it is easier to access.

AlexanderKaran:
@albertogasparin Added state to `replace`, `pushTo` and `replaceTo`. I also added state to the `Link` and changed it so we are not copying state and just using it from `location`. Due to the state being in location, I am not sure the extra setup and changes to the `useRouter` are needed as the four other functions get tested for a return in state and location is in `useRouter`

albertogasparin:
Looks great! I have one last ask: can you mention it in the docs? So it's clear that those APIs and Link accept `state` and you can find the value inside `location`

AlexanderKaran:
@albertogasparin Added an example and updated Link props, let me know if you need more

albertogasparin (approved):

PR #221, opened 2024-01-31
React 18 upgrade

AlexanderKaran:
First pass at React18 upgrade.

- Upgraded React packages
- Upgraded testing packages
- Upgraded tests to deal with changes


src/ui/link/utils/handle-navigation.tsx
```diff
@@ -24,15 +24,15 @@ export const handleNavigation = (
   event: any,
   { onClick, target, replace, routerActions, href, to, state }: LinkPressArgs
 ): void => {
-  if (isKeyboardEvent(event) && event.keyCode !== 13) {
+  if (isKeyboardEvent(event) && event.key === 'Enter') {
```
albertogasparin, line 27: This was a `!==` !

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x, 18.x]
+        node-version: [16.x, 18.x, 20.x]
```
albertogasparin, line 18: We can remove 16 and 18 then, no need to test on so many node versions...

package.json
```diff
@@ -128,17 +128,17 @@
...
   "peerDependencies": {
     "@babel/runtime": "^7",
     "history": "^4 || ^5",
-    "react": "^16.8.0",
-    "react-dom": "^16.8.0"
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
   },
   "engines": {
     "node": ">=10.0"
```
albertogasparin, line 144: Maybe this node should be updated too (to v16 at least)

src/controllers/router-store/test.tsx
```diff
@@ -1,4 +1,5 @@
 import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
```
albertogasparin, line 2: Why not importing this in the setup file? So we don't get strange errors if we move tests around

albertogasparin (changes requested):

src/controllers/router-store/test.tsx
AlexanderKaran: I tried moving in my previous PR but the TS compiler starts complaining and no matter what I tried it would not satisfy the error.

package.json
```diff
@@ -128,19 +128,19 @@
...
     "typescript": "^4.7.4",
     "webpack": "^4.44.1",
     "webpack-dev-server": "^4.11.1"
   },
   "peerDependencies": {
     "@babel/runtime": "^7",
     "history": "^4 || ^5",
-    "react": "^16.8.0",
-    "react-dom": "^16.8.0"
+    "react": "^18.2.0",
```
sgolovenko, line 140: should that be `"react": "^16.8.0 || ^18.2.0"` to be compatible with react16 in Jira?

src/controllers/router-store/test.tsx
AlexanderKaran: Bigger significant changes are needed to the config, and TSConfig is required. Basing this on all the simple changes I tried. 

Might be better in another PR as this one already has a lot of changes

src/controllers/use-query-param/test.tsx
```diff
@@ -272,16 +265,16 @@ describe('useQueryParam()', () => {
 
     const { storeState, actions } = getRouterStore();
 
-    actions.push('/projects/123/board/456?foo=hello&bar=world');
-    await nextTick();
+    act(() => {
```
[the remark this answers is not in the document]
AlexanderKaran: Were not using `userEvent` here so we need to manually use act due to waiting for re-renders and or any other background changes.

package.json
AlexanderKaran: Done, added types as well

sgolovenko (approved):

albertogasparin (approved):

Issue #223, opened 2024-02-29
Hydration Example Is Broken


AlexanderKaran:
The Hydration example page is broken and only returns a blank page. Even if I roll back, React18 changes, it still comes up as blank.

To replicate:

- Run npm start
- Go http://localhost:8080/hydration

PR #224, opened 2024-02-29
Fixed bug causing routes not to render on change

AlexanderKaran:
Fixes: https://github.com/atlassian-labs/react-resource-router/pull/224

- Fixed broken `useEffect` that was stopping correct rendering in `<StrictMode>`
- Updated examples to use the new React18 rendering method
- Add strict mode options to the integration tests

I did notice that the Hydration example is broken (even before the changes). I logged a bug for this: https://github.com/atlassian-labs/react-resource-router/issues/223

### DIFF
--- a/examples/basic-routing/index.tsx
+++ b/examples/basic-routing/index.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render } from 'react-dom';
+import { createRoot } from 'react-dom/client';
 
 import { About } from './about';
 import { Home } from './home';
@@ -42,4 +42,9 @@ const App = () => {
   );
 };
 
-render(<App />, document.getElementById('root'));
+const container = document.getElementById('root');
+if (!container)
+  throw new Error('No root element found to render basic routing example');
+
+const root = createRoot(container);
+root.render(<App />);
--- a/src/controllers/router/index.tsx
+++ b/src/controllers/router/index.tsx
@@ -15,13 +15,13 @@ export const Router = ({
   onPrefetch,
   routes,
 }: RouterProps) => {
-  useEffect(() => {
-    const { unlisten } = getRouterState();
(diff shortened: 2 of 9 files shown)


examples/basic-routing/index.tsx
```diff
@@ -42,4 +42,9 @@ const App = () => {
   );
 };
 
-render(<App />, document.getElementById('root'));
+const container = document.getElementById('root');
+if (!container)
+  throw new Error('No root element found to render basic routing example');
+
+const root = createRoot(container);
+root.render(<App />);
```
pancaspe87, line 50: How are we testing if it works with [StrictMode](https://react.dev/reference/react/StrictMode) ?

AlexanderKaran: See integration tests file

pancaspe87: yeahhh, I was looking at it now

pancaspe87 (approved):

src/controllers/router/index.tsx
```diff
@@ -15,13 +15,13 @@ export const Router = ({
   onPrefetch,
   routes,
 }: RouterProps) => {
-  useEffect(() => {
-    const { unlisten } = getRouterState();
+  const { unlisten } = getRouterState();
```
albertogasparin, line 18: Fine, even if I think the problem is the use of `getRouterState` which is bad (as out of react control flow)

albertogasparin (approved):

pancaspe87 (approved):

JohnsonChauAtlassian:
Cloned locally and confirmed that it works with/without strict mode in R18 concurrent mode for basic-routing example

JohnsonChauAtlassian (approved):

albertogasparin (approved):

Issue #225, opened 2024-03-04
Routing for new tabs or windows are not recognising `basePath`


tpuric:
Given `basePath` is enabled.

When attempting to open a link in a new tab (or window) via a `ctrl`/`cmd` click or right click new tab (or window), the new location omits the `basePath` prefix.

This behaviour is apparent in the `basic-routing` example.

E.g.
When on `http://localhost:8081/basic-routing/about`, opening the `Go to home` link in a new tab (or window) will lead to `http://localhost:8081/`.

tpuric:
I have also verified no regression. It’s had this behaviour since conception

tpuric:
I have explored this a little further and have noticed that it is only impacted for Links that provide `string` types of `to`. This behaviour is not shown when a `Route` is provided instead.

The `Route` path is verified by the following test
```typescript
    it('should render the correct link', () => {
      renderInRouter(
        'my link',
        {
          to: route,
          params: { id: '1' },
          query: { foo: 'bar' },
        },
        '/base'
      );
      expect(screen.getByRole('link', { name: 'my link' })).toHaveAttribute(
        'href',
        '/base/my-page/1?foo=bar'
      );
    });
```

I have constructed the following test which identifies the failure case for this issue.
```typescript
  it('should support the `to` prop with basePath', () => {
    renderInRouter(
      'my link',
      {
        to: newPath,
      },
      '/base'
    );
    const linkElement = screen.getByRole('link', { name: 'my link' });
    expect(linkElement).toHaveAttribute('href', `/base${newPath}`);
  });
```

```
    Expected the element to have attribute:
      href="/base/my-new-path"
    Received:
      href="/my-new-path"
```

PR #226, opened 2024-03-04
Supported basePath for to prop when passed as string. Resolving acces…

tpuric:
Resolves https://github.com/atlassian-labs/react-resource-router/issues/225

Supported basePath for to prop when passed as string.
Resolves accessibility issues for new tabs/window interactions.

### DIFF
--- a/src/ui/link/index.tsx
+++ b/src/ui/link/index.tsx
@@ -71,7 +71,7 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
               generateLocationFromPath(route.path, routeAttributes)
             )) ||
           ''
-        : to;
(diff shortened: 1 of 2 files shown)


src/ui/link/index.tsx
```diff
@@ -71,7 +71,7 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
               generateLocationFromPath(route.path, routeAttributes)
             )) ||
           ''
-        : to;
+        : `${routeAttributes.basePath}${to}`;
```
tpuric, line 74: I'm not sure if there's a better way to handle this. It's essentially the same logic as `generateLocationFromPath` except this should already by a generated path, thus no transformations.

PR #227, opened 2024-03-04
Supported basePath for to prop when passed as string. Resolving acces…

tpuric:
Resolves https://github.com/atlassian-labs/react-resource-router/issues/225

Supported basePath for to prop when passed as string.
Resolves accessibility issues for new tabs/window interactions.

Test added:
* should support the `to` prop with basePath
* should push history with correct link given basePath

basePath

### DIFF
--- a/src/ui/link/index.tsx
+++ b/src/ui/link/index.tsx
@@ -72,6 +72,16 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
             )) ||
           ''
         : to;
+    const hrefDestination =
+      href != null
+        ? href
+        : typeof to !== 'string'
+        ? (route &&
+            createPath(
+              generateLocationFromPath(route.path, routeAttributes)
+            )) ||
+          ''
+        : `${routeAttributes.basePath}${to}`;
 
     const triggerPrefetch = useCallback(() => {
       // ignore if async route not ready yet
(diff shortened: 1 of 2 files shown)


src/ui/link/index.tsx
```diff
@@ -72,6 +72,16 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
...
+    const hrefDestination =
+      href != null
+        ? href
+        : typeof to !== 'string'
+        ? (route &&
+            createPath(
+              generateLocationFromPath(route.path, routeAttributes)
+            )) ||
+          ''
+        : `${routeAttributes.basePath}${to}`;
```
tpuric, line 84: yeah so this is obviously hacked in, looking for input on a better way of handling this.

Essentially the problem here is that `linkDestination` is used for both router actions and anchor props (href). Router actions automatically prepends basePath whilst anchor props do not, hence why we are in the current state.

This example splits the logic so they are handled separately. The approach is very naive, please provide suggestions. Alternatively, we could look at converting string typed `to`'s to routes?

Anyways, looking for feedback, thanks.

albertogasparin: We already calculate the href above, don't need to recalculate everything again. We just need to understand if we need to add basepath.

```js
const staticBasePath = href == null && typeof to === 'string' ? routeAttributes.basePath : ''
// ...
validLinkType,
  {
  ...rest,
  href: staticBasePath + linkDestination,
```

tpuric: I've applied these changes, thanks @albertogasparin

pancaspe87 (approved):

PR #241, opened 2024-05-27
 Document customizing getKey and getData fetching

DmitryGonchar:
e.g. to pass a different route parameter value to them.

These clarifications are result of a pairing session with @albertogasparin


docs/resources/creation.md
```diff
@@ -17,9 +17,9 @@ export const userProfileResource = createResource({
...
+| `getKey`        | `(routerContext, customContext) => string`               | The return value of this function is used to identify this resource within the `type` namespace. The function itself is supplied with `routerContext` and `customContext` so that the composition of keys can use this data if required. `customContext` comes from the top level Router component, therefore changing it will cause a re-render of almost everything in your React app. You may want to pass a custom `routerContext` instead. How to do this - see the ["Accessing resource state for another route or url"](https://atlassian-labs.github.io/react-resource-router/#/resources/usage?id=accessing-resource-state-for-another-route-or-url) section. The router will fetch and cache data for that route then.                                                                                       |
```
liamqma, line 20: I don't quite understand this sentence in the context of `getKey` function.
> therefore changing it will cause a re-render of almost everything in your React app

`customContext` is immutable within `getKey` function scope, isn't it?

DmitryGonchar: That's what Alberto said, can you please clarify with him, and make edits if necessary?

Afaiu, if we put a new prop into `customContext`, it means that all child components of Router will get re-evaluated. And it was not recommended for that reason. But yes, adding more specific details here'd be helpful, I just don't have them :)

albertogasparin: Maybe the confusion is because describing side effects of changing customContext while we talk about resources is misleading. Maybe something like this is easer to grasp?

> The function itself is supplied with `routerContext` and `customContext` so that the composition of keys can use this data if required. `customContext` comes from the top level Router component and therefore is value is equal across resources. `routerContext` contains the current route and location data, which comes from the current URL or from some manually provided value (see the ["Accessing resource state for another route or url"](https://atlassian-labs.github.io/react-resource-router/#/resources/usage?id=accessing-resource-state-for-another-route-or-url) section). The router will fetch and cache data for that route then.

DmitryGonchar: Lgtm. @liamqma is Alberto's version clearer? If yes, I can update the PR

liamqma: Make senses. Thank you @albertogasparin and @DmitryGonchar

liamqma (approved):

PR #242, opened 2024-06-14
Updated query params docs

pancaspe87:
Screenshot 2024-06-14 at 3 00 13 PM

A very minor update, devs are not aware of this option

### DIFF
--- a/docs/router/configuration.md
+++ b/docs/router/configuration.md
@@ -48,7 +48,7 @@ export const routes = [
 ];
 ```
 
-You can also mark a path param as optional by using the `?` suffix. For example, if you want to make the `userId` param optional, you would do so like this `'/user/:userId?'`.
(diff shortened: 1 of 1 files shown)


at-nathan (approved):

docs/router/configuration.md
```diff
@@ -48,7 +48,7 @@ export const routes = [
 ];
 ```
 
-You can also mark a path param as optional by using the `?` suffix. For example, if you want to make the `userId` param optional, you would do so like this `'/user/:userId?'`.
+You can also mark a path param as optional by using the `?` suffix. For example, if you want to make the `userId` param optional, you would do so like this `'/user/:userId?'`. This also applies to query parameters, e.g `query: ['foo', 'baz?']`.
```
ParkerRobinson, line 51: Theres also a comment on InvariantRoute that explains what 'query' is and doesn't specify you can create optional query params.

JakeLane (approved):

PR #245, opened 2024-11-04
Bumped path-to-regexp

AlexanderKaran:


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "react-resource-router",
-  "version": "0.28.0",
+  "version": "0.29.0",
```
JakeLane, line 3: Missing `package.json` change?

JakeLane (changes requested):

package-lock.json
AlexanderKaran: Package.json is already at "version": "0.29.0",

JakeLane: I guess this probably means changesets don't run the package manager correctly on bump, I suppose that's fine since it's local to the repo

JakeLane (approved):

PR #247, opened 2024-11-05
Add no-cycle eslint rule

nicholas-l:
We are trying to mock our imports of react-resource-router using snippet below, however due to the cyclic import of react-resource-router, we are getting a "Cannot read properties of undefined (reading 'ResourceDependencyError')". I believe this is due to the cyclic imports that means when Jest tries to import the actual module, and then when the submodules in react-resource-router imports the root index file, this file points to the Jest mock which has not been fully set up. This causes an property of undefined error.

```js
jest.mock('react-resource-router', () => ({
   ...jest.requireActual('react-resource-router'),
   useRouter: jest.fn(),
});
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -45,7 +45,7 @@ module.exports = {
     'react-hooks/exhaustive-deps': 'warn',
     'react-hooks/rules-of-hooks': 'error',
     semi: 'off',
-    'import/no-cycle': 'warn',
+    'import/no-cycle': 'error',
   },
   overrides: [
     {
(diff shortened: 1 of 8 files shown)


.eslintrc.js
```diff
@@ -45,6 +45,7 @@ module.exports = {
     'react-hooks/exhaustive-deps': 'warn',
     'react-hooks/rules-of-hooks': 'error',
     semi: 'off',
+    'import/no-cycle': 'warn',
```
JakeLane, line 48: make this an error instead?

JakeLane (approved):

PR #248, opened 2024-12-16
Support basePath for relative href prop

upreeti:
We are using the `basePath` prop for navigation in our application and for our side navigation we are using the MenuLinkItem which takes an `href` prop. Because of this, we did not see the `basePath` being added to the route when hovering over the side navigation. 

I found [this PR](https://github.com/atlassian-labs/react-resource-router/pull/227/files) which prefixes basePath when using the `to` prop, and I was wondering if this could be extended to relative hrefs as well. We added a patch fix for this in our code, but I created this PR to upstream this change of prefixing to relative `href` paths, after reaching out to @pancaspe87

### DIFF
--- a/src/ui/link/index.tsx
+++ b/src/ui/link/index.tsx
@@ -72,7 +72,7 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
               )) ||
             ''
           : to;
-    const IS_ABSOLUTE_LINK_REGEX = /^(?:(http|https):\/\/)/;
+    const IS_ABSOLUTE_LINK_REGEX = /^((?:(http|https):\/\/)|\/\/)/;
     const staticBasePath =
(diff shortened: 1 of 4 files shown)


pancaspe87 (approved):

src/ui/link/index.tsx
```diff
@@ -72,8 +72,12 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
               )) ||
             ''
           : to;
+    const IS_ABSOLUTE_LINK_REGEX = /^((?:(http|https):\/\/)|\/\/)/;
```
liamqma, line 75: The code change itself looks good. However, I wonder if href should be prefixed.
For `to`, it makes sense because `to` must be one of routes. 
But for `href`, it's basically any URL. Should we prefix at the product side?

```
const { getBasePath } = useRouterActions();
const basePath = getBasePath();

const href = `${getBasePath}/something`;
```

liamqma: Is it too many places we need to prefix if it's done at the product level?

liamqma (approved):

PR #250, opened 2025-01-02
order independent routing

liamqma:
This PR adds order-independent matchRoute.

### DIFF
--- a/src/common/utils/match-route-order-independent/index.tsx
+++ b/src/common/utils/match-route-order-independent/index.tsx
@@ -1,0 +1,326 @@
+import { pathToRegexp } from 'path-to-regexp';
+import { qs } from 'url-parse';
+
+import { Query, Routes, Route } from '../../types';
+import execRouteMatching from '../match-route/exec-route-matching';
+import matchQuery from '../match-route/matchQuery';
+
+import { type Tree, Node, treeify } from './tree';
+import { matchRouteCache } from './utils';
+
+function pushOrUnshiftByCaptureGroup(arr: Node[], node: Node) {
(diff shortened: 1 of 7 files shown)


vykimnguyen (approved):

src/common/utils/match-route-order-independent/index.tsx
```diff
@@ -0,0 +1,326 @@
+import { pathToRegexp } from 'path-to-regexp';
```
at-nathan, line 1: Should we use `.ts` file extension for files in this directory?

src/common/utils/match-route-order-independent/utils.tsx
```diff
@@ -0,0 +1,69 @@
+import type { Match, Query } from '../../types.tsx';
+
+const hasOwnProperty = Object.prototype.hasOwnProperty;
+const MAX_CACHE_SIZE = 1000;
+
+function shallowEqual(objA: any, objB: any) {
```
at-nathan, line 6: Is this a copied from somewhere? I remember seeing something similar already in the repo.

src/common/utils/match-route-order-independent/index.tsx
liamqma: ah, good point. Let me change it to `.ts`.

src/common/utils/match-route-order-independent/utils.tsx
liamqma: Yes, it's a duplicate. I copied it because I want to keep the cache separate between the existing matchRoute and the order-independent matchRoute.
Thinking about it, I can create a factory function to avoid duplicated code. Let me do that.

PR #252, opened 2025-02-27
Supporting navigation to / and basePath when basePath is defined

upreeti:
The current issue is that when `basePath` is defined (e.g. `/base`), even when we pass in `/` and `/base` as routes to RRR, navigating to `url.com/` or `url.com/base` results in a 404, because 

- `/` becomes `/base/` and 
- `/base` becomes `/base/base`

when trying to find matching routes, and so now no routes match `/` or `/base`.  This PR is making an update so that when the user is navigating to `/` or `/base` the base path is ignored and they become valid routes to match on. `/base/` is still a valid route

### DIFF
--- a/.changeset/ninety-penguins-speak.md
+++ b/.changeset/ninety-penguins-speak.md
@@ -1,5 +1,5 @@
 ---
-"react-resource-router": patch
+"react-resource-router": minor
 ---
 
 Supporting navigation to / and basePath when basePath is defined
--- a/.changeset/ninety-penguins-speak.md
+++ b/.changeset/ninety-penguins-speak.md
@@ -1,0 +1,5 @@
+---
+"react-resource-router": patch
+---
+
+Supporting navigation to / and basePath when basePath is defined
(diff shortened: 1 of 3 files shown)


src/common/utils/match-route/index.ts
```diff
@@ -43,10 +45,15 @@ const matchRoute = <T extends Route | InvariantRoute>(
...
       queryParamObject,
-      basePath
+      updatedBasePath
     );
     if (matchedRoute) {
-      matchRouteCache.set(pathname, queryParamObject, basePath, matchedRoute);
+      matchRouteCache.set(
+        pathname,
+        queryParamObject,
+        updatedBasePath,
```
upreeti, line 54: formatting

liamqma (approved):

.changeset/ninety-penguins-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-resource-router": patch
```
liamqma, line 2: To be safe it needs to be a minor I think. Although we've fixed a bug, we've also introduced a new behaviour - allow `/` if basePath is set.

upreeti: okay will update

liamqma (approved):

PR #256, opened 2025-06-03
pass onKeyDown to Link component

joeyslater:
current state:
- onKeyDown points to `onLinkPressed` -> `handleNavigation` which swallows the original onKeyDown.
- onClick is properly passed to `onLinkPressed` -> `handleNavigation`

pr state: 
- passes onKeyDown to `handleNavigation` (which one could argue we should handle in the Link component itself, but that was a little bit bigger of a refactor - which I can do if the team wants); when the key is not Enter, the onKeyDown will be fired.

### DIFF
--- a/src/ui/link/index.tsx
+++ b/src/ui/link/index.tsx
@@ -31,6 +31,7 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
       href = undefined,
       to = undefined,
       onClick = undefined,
+      onKeyDown = undefined,
       onMouseEnter = undefined,
       onMouseLeave = undefined,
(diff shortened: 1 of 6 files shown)


src/ui/link/index.tsx
```diff
@@ -103,6 +104,7 @@ const Link = forwardRef<HTMLButtonElement | HTMLAnchorElement, LinkProps>(
     const handleLinkPress = (e: MouseEvent | KeyboardEvent) =>
```
tomgasson, line 104: nit: Given this is immediately split into keyboard vs non-keyboard inside `handleNavigation`, might as well just fully split this into two different things

